PR #2, opened 2023-03-30
Basic proof of concept

richfitz:
This PR demonstrates the core idea for the most simple case I can come up with - just running the minimal example. I've made two versions here, one that is explicit about inputs and outputs and one that is implicit. There's lots of stuff I resisted adding here because it's more controversial - even just doing parameters would be nice to do in a separate PR I think.

The basic pattern though will remain the same - we'll have two versions of each function that the user might call from within their orderly.R file - one that is "static"; that one we'll try and poke at when we read the file before running and we can enforce many existing orderly rules this way. Later we can make a "strict mode" that requires that all calls are done this way. The second version is the one that gets actually run when the report runs - that accumulates data into a field within the packet state.

There are a few bits here that we might want to think about - in particular when reading we might want to limit which orderly3 calls are allowed in the file. That's fairly easy to do though and can be done later.

Note that these can be run from within their source directories, just like any other R script. As we get more complicated some more work will be required to make that actually behave well I suspect


tests/testthat/test-read.R
```diff
@@ -0,0 +1,30 @@
...
+
+
+test_that("Can read string literals from expressions", {
+  expect_equal(static_character_vector(quote("x")), "x")
+  expect_equal(static_character_vector(quote(c("x"))), "x")
+  expect_equal(static_character_vector(quote(c("x", "y"))), c("x", "y"))
+  expect_equal(static_character_vector(quote(c("x", c("y", "z")))),
+               c("x", "y", "z"))
+
+  expect_null(static_character_vector(quote(a)))
```
r-ash, line 23: So we can't pass a variable into `orderly_resource` or `orderly_artefact`? That makes sense, but I wonder if need to improve the error messaging if you try and do that? Perhaps something for later, if you do e.g.

```
my_file <- "data.csv"
orderly3::orderly_resource(my_file)
```

when trying to run this with `orderly_run` it gives
```
>   id <- orderly_run("explicit", root = path, envir = env)
Error in static_character_vector(files) : object 'my_file' not found
```

R/run.R
```diff
@@ -0,0 +1,126 @@
...
+    outpack::outpack_packet_run("orderly.R", envir, packet = p)
+    check_produced_artefacts(path, p$orderly3$artefacts)
+    custom_metadata_json <- to_json(custom_metadata(p$orderly3))
+    outpack::outpack_packet_add_custom("orderly", custom_metadata_json,
+                                       custom_metadata_schema(), packet = p)
+    outpack::outpack_packet_end(p)
+    unlink(path, recursive = TRUE)
+  }, error = function(e) {
+    ## Eventually fail nicely here with mrc-3379
+    outpack::outpack_packet_cancel(p)
```
r-ash, line 91: This will fail with confusing error if `orderly_packet_start` doesn't exist, should we not call this if `p` doesn't exist?

tests/testthat/helper-orderly.R
```diff
@@ -0,0 +1,26 @@
...
+  withr::defer_parent(unlink(tmp, recursive = TRUE))
+  orderly_init(tmp)
+
+  config <- character()
+
+  if ("global" %in% examples) {
+    config <- c(config,
+                "global_resources: global")
+    fs::dir_create(file.path(tmp, "global"))
+    fs::file_copy("examples/minimal/data.csv", file.path(tmp, "global"))
```
r-ash, line 17: Can we wrap in `testthat::test_path` so then the tests will work interactively without changing your working directory?

```suggestion
    fs::file_copy(testthat::test_path("examples/minimal/data.csv"), file.path(tmp, "global"))
```

tests/testthat/helper-orderly.R
```diff
@@ -0,0 +1,26 @@
...
+  }
+
+  writeLines(config, file.path(tmp, "orderly_config.yml"))
+  fs::dir_create(file.path(tmp, "src"))
+  for (i in examples) {
+    fs::dir_copy(file.path("examples", i), file.path(tmp, "src"))
```
r-ash, line 23:
```suggestion
    fs::dir_copy(testthat::test_path("examples", i), file.path(tmp, "src"))
```

r-ash (commented):
This looks good, definitely some weird bits but I suppose working with the language itself is a bit weird. Couple of comments, possibly for later

R/util.R
```diff
@@ -6,3 +6,59 @@
 is_directory <- function(x) {
   file.info(x, extra_cols = FALSE)$isdir
 }
+
+
+is_call <- function(x, name) {
+  is.recursive(x) && is.name(x[[1]]) && as.character(x[[1]]) == name
```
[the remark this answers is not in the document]
richfitz:
```
is.call(list(1, 2)) # FALSE
is.recursive(list(1, 2)) # TRUE
```

I tend to use is.recursive here as the more general one but also because in this context it reads better (something we might recurse into). Note that `is_call` here as I'm writing it here is really ("is specific call")

tests/testthat/test-read.R
richfitz: thanks, this is a bug, we should be able to do that. I'll have  look why that's not getting the right environment

R/run.R
richfitz: good point, do you have an example where it fails already? otherwise I can probably contrive one

tests/testthat/helper-orderly.R
richfitz: yeah, absolutely

R/util.R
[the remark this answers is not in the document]
r-ash: Thanks! Actually looking at docs again `is.recursive returns TRUE if x has a recursive (list-like)` is a good hint

R/run.R
r-ash: Not a good one, in fact I only saw that because I was running with an old version of `outpack` which was expecting different args to `orderly_packet_start`. Would be fairly easy to mock it

r-ash: O wait we can't mock functions from other packages anymore can we :cry:

richfitz: yes, that works fine. I'll set up a test with that

tests/testthat/helper-orderly.R
richfitz: Ugh, turns out that with this you _have_ to be in the root, I thought it worked both ways?

R/run.R
richfitz: Actually just moving it outside the withCallingHandlers is the simplest thing and always correct

r-ash (approved):

richfitz:
@hillalex ; merging this into the `epic-poc` branch so you review incrementally after easter

PR #3, opened 2023-04-03
Basic parameter support

richfitz:
This is basically the logic (and mostly the code) from orderly (v1 and 2) to support parameters. I think this is needed before getting to dependencies really.

There's quite a different path here interactively, but I've shared code where I can. I don't think there's actually very much interesting in this one compared with the previous PR tbh.


R/read.R
```diff
@@ -28,6 +29,20 @@ orderly_read_r <- function(path) {
...
+    ## TODO: once things are working, check for no duplicate parameter
+    ## names across multiple calls; leaving this assertion in for
+    ## now. Reporting on that well suggests that we should be able to
+    ## record the line numbers when we pass off through the file;
+    ## that's not too hard to do really, I think we can do that with
+    ## getSrcref or similar.
+    ##
+    ## TODO: we should disallow use of the parameter function in any
+    ## deeper expression (at least in strict mode), worth searching
+    ## for that and setting up that check here
```
r-ash, line 42: Do you want to address any of these or leave them until later?

r-ash (approved):
Looks good, one small comment about the TODOs

R/read.R
richfitz: Later, both of these are fairly complex and we'll hit similar issues in other parts I think

R/metadata.R
```diff
@@ -1,3 +1,43 @@
...
+##'
+##' @param ... Any number of parameters
+##'
+##' @return Undefined
+##'
+##' @export
+orderly_parameters <- function(...) {
+  p <- get_active_packet()
+  if (is.null(p)) {
+    ## Here, we might prompt for the presence of parameters in the
```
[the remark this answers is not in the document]
richfitz: https://github.com/mrc-ide/orderly3/pull/9

PR #5, opened 2023-04-04
Basic support for dependencies

richfitz:
As with the previous PRs, mostly fairly straightforward I hope. Major issues are:

* detection of a "safe" location to copy files into. We need this for global resources too.
* we can probably be more relaxed about what we pull from the arguments to the dependency function, particularly handling around the query parameter, but that can be done later I think


tests/testthat/examples/depends/orderly.R
```diff
@@ -0,0 +1,2 @@
+orderly3::orderly_depends("explicit", "latest", c(graph.png = "mygraph.png"))
+orderly3::orderly_artefact("Final plot", "graph.png")
```
r-ash, line 2: Can we add a test case where the name of the dependency is different from the artefact? Ran into an issue with this in orderly2 which is what https://github.com/vimc/orderly2/pull/8 is addressing. Could also do that in a separate PR.

r-ash (approved):
Looks good, one comment about an additional test but could also leave that for a separate PR

tests/testthat/examples/depends/orderly.R
richfitz: In this example the dependency is being renamed, but there's no "unexpected files" check yet here (it'll be a bit harder to add). Or is it something else that's missing in the orderly2 test?

r-ash: I mean the name of the dependency after being pulled into this report run is different from the name of the artefact. They are the same here as you say. Though actually if we have no unexpected files check here then comment above is not relevant!

NAMESPACE
```diff
@@ -1,6 +1,7 @@
 # Generated by roxygen2: do not edit by hand
 
 export(orderly_artefact)
+export(orderly_depends)
```
hillalex, line 4: I guess the most consistent with the other function names would be `orderly_dependency`

R/metadata.R
```diff
@@ -120,6 +120,79 @@ static_orderly_artefact <- function(args) {
...
+##'   want a more complex query here though.
+##'
+##' @param use A named character vector of filenames to copy from the
+##'   upstream packet. The name corresponds to the destination name,
+##'   so c(here.csv = "there.csv") will take the upstream file
+##'   `there.csv` and copy it over as `here.csv`.
+##'
+##' @return Undefined
+##' @export
+orderly_depends <- function(name, query, use) {
```
hillalex, line 140: Feels like it would be nice if you could just do `function(name, query, ...)` where `...` is the named file arguments.

richfitz: I agree, but here's the pair of issues issue with that:

First, In the case where you want to compute the filenames passing in dots is hard as R does not have a spread/splat operator. This will be an issue for Pete's case where we are pulling in things from different regions and renaming them as what we can do with an explicit `use` argument is:

```r
for (r in regions) {
  query <- rlang::expr(parameter:region == !!r)
  use <- setNames("data.rds", file.path("data", paste0(r, ".rds")))
  orderly_depends("data", query, use)
}
```

Second, you can compromise by adding a get collisions with names, so we could write:

```
orderly_depends <- function(name, query, ..., use = c(...)) {
}
```

or by carefully processing `...`, but this requires some care with any analysis of the contents in `static_orderly_dependency` I think.

I had a look at sorting this out but it looks complicated enough that I'd rather kick it down the road a bit

richfitz: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-4173

PR #8, opened 2023-04-12
Add support for plugins

richfitz:
This PR largely follows the same ideas as in orderly2 (see https://github.com/vimc/orderly2/pull/6) with a bunch of code pulled in verbatim, but tweaked to work with the new idea of things being interactive.

I've simplified the configuration structure in `orderly_config.yml` a bit from orderly2, now there's just one bit to add which is much easier to understand.

I can add a vignette here that explains things, but in order to keep this PR more manageable I've posted a comment with a render of it below.

You can see a nontrivial example here, with the updated version of the orderly db plugin: https://github.com/mrc-ide/orderly3.db/pull/1


r-ash (approved):
This looks great to me!

R/plugin.R
```diff
@@ -0,0 +1,214 @@
...
+orderly_plugin_environment <- function(name) {
+  frames <- sys.frames()
+  for (i in seq_along(frames)[-1]) {
+    if (environmentName(parent.env(frames[[i]])) == name) {
+      return(frames[[i - 1]])
+    }
+  }
+  ## This error should never surface if the plugin is configured correctly
+  stop("Could not determine plugin calling environment safely - please report")
+}
```
r-ash, line 207: This will be useful actually when trying to get subqueries working with `orderly_depends`. We want to be able to do something like

```
orderly3::orderly_depends("depends", "latest",
                          c(depends_graph.png = "graph.png"),
                          as = "depends")
orderly3::orderly_depends("explicit", "usedby({depends})",
                          c(explicit_graph.png = "mygraph.png"))
```

which is fine if running via `orderly_run` as we can use the active packet environment but if running interactively we're going to need to determine the environment like this. Any thoughts on if we should try and make this generic or just add a similar function?

r-ash: Oh actually maybe we just need the `parent.frame` in the context I am talking about!

richfitz: I was discussing this with Alex - I think that the whole context thing might want exposing more internally. It would be useful for the global dependencies too

R/metadata.R
```diff
@@ -149,7 +156,8 @@ orderly_depends <- function(name, query, use) {
     path <- getwd()
     root <- detect_orderly_interactive_path(path)
     env <- parent.frame()
-    id <- outpack::outpack_query(query, env, name = name,
+    parameters <- current_orderly_parameters(path, env)
```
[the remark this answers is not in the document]
hillalex: Ah yeah, I see

hillalex (approved):

PR #13, opened 2023-04-25
Support interpolating environment variables into config

richfitz:
As used in orderly, needed for practical use of the database plugin


r-ash (approved):
Looks good!

PR #16, opened 2023-05-02
Strict mode / sensible handling of resources

richfitz:
This PR is a fixed version of #14:

> This PR pulls together the basic idea for "strict mode", in which we start on some support that VIMC folk will expect. This mostly concerns indicating when unexpected files are produced and encouraging the user to annotate as required. We take two different approaches here depending on the value of `orderly_strict_mode`:
> 
> * if enabled, then anything that is not declared gets warned about and we only copy files that are explicitly requested
> * if disabled, we copy everything except artefacts and then try and detect things that we might have copied in error (this is the case where the user has been developing locally and we copy over an artefact as if it was an input). We use a simple heuristic of "file was modified after copy based on size and modification time". Note that modification time resolution on windows is very poor.
> 
> It only makes sense to include `orderly_strict_mode` once, and at the top level (so that we work on the inferred value of it, not from the point that it is called). I've added in a check that ensures that no conditional calls etc are included and also applied this to `orderly_parameters` where we had a comment indicating that the same treatment applied.
> 
> Because the processing of the file was getting very loopy I've done this in a more efficient, but equivalent way.
> 
> Once this is in:
> 
> * we can start more reasonably on the equivalent of "orderly_develop_clean" which will be much safer in strict mode
> * we can look at how this could be enforced at a server level, or overridden
> * we might eventually break up some of the bits that go into "strict mode" into options to the call - I've foreshadowed this by that list with `enabled = TRUE` as the only element.

The issue with #14 as-is is that is exacerbates some issues with resources - because they're not really handled by the `orderly_resource` function they were incorrect in many places, including where they were computed or where they involved directories. This PR is probably no harder to review and is at least correct. But it is an overhaul of how resource copying works.

The PR is failing because posit package manager is down, it should be back up and running soon


r-ash (approved):
This also looks great. There are a few TODOs floating around, up to you if you want to leave them in or ticket up.

hillalex (approved):
I have read this and the functionality all looks good to me, but not sure I can offer much in the way of useful commentary!

richfitz:
> I have read this and the functionality all looks good to me, but not sure I can offer much in the way of useful commentary!

yeah, sorry about this one it's super churny and very fiddly...

PR #19, opened 2023-05-12
Cope better with failed scripts

richfitz:
This one, in contrast with the outpack support (https://github.com/mrc-ide/outpack/pull/56), is fairly straightforward:

* general refactoring of the main run function to expose two clear cleanup paths - the happy one and the unhappy one
* happy path has no logic changes
* unhappy path uses new outpack function to dump out metadata (see that PR for details) and will surface errors to the user as they happen (hard to test automatically because of the way the handerlers work)

### DIFF
--- a/tests/testthat/test-run.R
+++ b/tests/testthat/test-run.R
@@ -610,3 +610,25 @@ test_that("can enable logging at the packet level", {
                fixed = TRUE, all = FALSE)
   expect_match(res$output, "orderly3::orderly_artefact")
 })
+
+
+
+test_that("cope with failed run", {
+  path <- test_prepare_orderly_example("implicit")
+  prepend_lines(file.path(path, "src", "implicit", "orderly.R"),
+                'd <- read.csv("something.csv")')
+  err <- expect_error(
+    orderly_run("implicit", root = path),
+    "cannot open the connection",
+    class = "outpack_packet_run_error")
+  expect_length(dir(file.path(path, "archive", "implicit")), 0)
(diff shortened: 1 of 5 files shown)


tests/testthat/test-run.R
```diff
@@ -610,3 +610,25 @@ test_that("can enable logging at the packet level", {
...
+
+
+test_that("cope with failed run", {
+  path <- test_prepare_orderly_example("implicit")
+  prepend_lines(file.path(path, "src", "implicit", "orderly.R"),
+                'd <- read.csv("something.csv")')
+  err <- expect_error(
+    orderly_run("implicit", root = path),
+    "cannot open the connection",
+    class = "outpack_packet_run_error")
```
r-ash, line 623: Wow this made me realise the message when reading a file "cannot open file 'thing.csv': No such file or directory" is actually a warning instead of the error :fearful:. It's a bit of a shame we can't show that message when a user runs `orderly_run` here but perhaps it's hard to work around as that is specific to the error handling of read.csv. It is available in the `output` of the object

r-ash (approved):
This looks good too, one comment but not sure it is something we can reasonably work around

tests/testthat/test-run.R
richfitz: yeah, this is really annoying - it should be an error that is visible as part of the other error. Definitely not something that we should handle specifically. However, we could look at how something like callr shows some context on the screen at the point where the error is thrown?

r-ash: Yeah that sounds like a good plan

PR #21, opened 2023-06-05
vimc-7125: let outpack do the search for us

richfitz:
This PR lets outpack do the search for us (rather than orderly3 itself) when finding dependencies, which is necessary for safely including the query into the metadata.

* the `search_options` interface makes lots of possibilities available, generalising the old `use_draft` interface. I've added a big block of docs about this.
* most of the interesting changes are in `orderly_dependency` which should be fairly easy to read and certainly a little simpler. There are some improvements marked with a TODO that require a little outpack work, but we can come back to sort that out later I think
* the fallout in `orderly_run` is just passing an extrac argument around.
* the `orderly_run_info` function is hard to keep happy in this, and I don't think this ever worked well in orderly1, so happy to punt on this if that's ok.
* because we don't pass options directly into `orderly_dependency` we need some control over this interactively; see the little `orderly_interactive_set_search_options` function. I imagine we might change its name at some point, but it should get the job done.  See tests for use.

~Depends on https://github.com/mrc-ide/outpack/pull/72 and https://github.com/mrc-ide/outpack/pull/73 but otherwise ready for review - update branch pins before merge~


R/run.R
```diff
@@ -316,7 +379,7 @@ orderly_packet_cleanup_success <- function(p) {
 
 
 orderly_packet_cleanup_failure <- function(p) {
-  ignore_errors(plugin_run_cleanup(path, p$orderly3$config$plugins))
+  ignore_errors(plugin_run_cleanup(p$path, p$orderly3$config$plugins))
```
richfitz, line 382: This was broken before

r-ash (approved):
Looks good to me, one small typo I noticed but impl looks great

PR #25, opened 2023-07-06
Import all code from outpack

richfitz:
Massive update, but no sneaky features added.

This PR adds all the outpack code. Most outpack code has ended up in files that begin `R/outpack_` or `tests/testthat/test-outpack-` which feels ok. The NAMESPACE makes it fairly obvious how things should be used. Modules-within-package structures would make this nice and easy but that's probably never going to happen.

I did drop the case sensitive file_exists checks because outpack never had them and they failed really badly in windows here (c9e0d91 and mrc-4375 to restore)


r-ash (approved):
Had a look through the merged files and this is looking good

PR #26, opened 2023-07-07
Export orderly_init

richfitz:
As noted when we looked through, this PR exports orderly_init properly

I've not gone (yet) for harmonising the configurations though as we might want to keep them separate?


r-ash (approved):
Looks good, sounds fine to leave separate - especially if we want to have the option of splitting orderly and outpack again in the future

PR #27, opened 2023-07-07
Allow custom locations to be added

richfitz:
This is a bit ropey, and worth chatting about. In particular, I wonder if instead of "custom" type we might use something like the fully qualified driver name as the type? Not sure, really. The issue here is that we expect that other outpack implementations (e.g. a python or rust one) pointed at this archive need to know that they don't need to stress about the location entry.

Look at https://github.com/mrc-ide/outpack.sharepoint/pull/1 for the implementation of a custom driver that inspired this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.1
+Version: 1.99.2
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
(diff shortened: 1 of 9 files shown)


r-ash (approved):
Looks good, "custom" sounds good to me. Though I suppose if we add another "custom" type at some point which is supported in Rust or Python then we'll have to do some switching on the first arg to a custom. Instead of just checking the name itself. But I think that's probably fine?

richfitz:
yeah, I really don't know how this is going to go in practice with another implementation (that's really why I left that comment about future change there). I expect we'll need to move things around a bit

PR #28, opened 2023-07-10
Tidy behaviour when querying for boolean types

r-ash:
This PR will
* Update outpack_query so users can search for boolean with either `TRUE`, `FALSE` or `true`, `false`. We no longer support `T`, `"TRUE"`, `1` 
* Checks that the type of query parameter matches the type of the parameter in the metadata

This updates https://github.com/mrc-ide/outpack/pull/74 for orderly2 repo and addresses the comments there.


R/query_search.R
```diff
@@ -231,8 +231,14 @@ query_eval_test_binary <- function(op, a, b) {
   if (length(a) == 0 || length(b) == 0) {
     return(logical(0))
   }
-  vlapply(Map(function(a, b) !is.null(a) && !is.null(b) && op(a, b),
-              a, b, USE.NAMES = FALSE),
+  is_equal <- function(a, b) {
+    !is.null(a) && !is.null(b) &&
+      (is.numeric(a) && is.numeric(b) ||
+         is.logical(a) && is.logical(b) ||
+         is.character(a) && is.character(b)) &&
```
richfitz, line 238: `storage.mode(a) == storage.mode(b)
```suggestion
      storage.mode(a) == storage.mode(b) &&
```

R/query_search.R
```diff
@@ -231,8 +231,14 @@ query_eval_test_binary <- function(op, a, b) {
   if (length(a) == 0 || length(b) == 0) {
     return(logical(0))
   }
-  vlapply(Map(function(a, b) !is.null(a) && !is.null(b) && op(a, b),
-              a, b, USE.NAMES = FALSE),
+  is_equal <- function(a, b) {
```
richfitz, line 234: consider moving this into utils, makes it easier to test and does not need to be an inner function

richfitz (changes requested):

R/query_search.R
r-ash: Is there a way we can make this work for integer and numeric types?

```
> storage.mode(1L) == storage.mode(1)
[1] FALSE
```

richfitz: good point. This should work:

```
same_type <- function(a, b) {
  (is.numeric(a) && is.numeric(b)) || (storage.mode(a) == storage.mode(b))
}
```

R/query.R
```diff
@@ -284,6 +284,22 @@ is_id_lookup <- function(expr) {
...
+  truthy <- c(quote(TRUE), quote(FALSE), quote(true), quote(false))
+  bool <- vlapply(truthy, function(x) x == expr)
+  sum(bool) == 1
+}
+
+as_logical <- function(expr) {
+  if (expr == quote(true)) {
+    ret <- quote(TRUE)
+  } else if (expr == quote(false)) {
+    ret <- quote(FALSE)
```
richfitz, line 297:
```suggestion
    ret <- TRUE
  } else if (expr == quote(false)) {
    ret <- FALSE
```

no need to quote these, it  has no effect (`identical(TRUE, quote(TRUE))` is `TRUE`)

R/query.R
```diff
@@ -284,6 +284,22 @@ is_id_lookup <- function(expr) {
   expr$type == "lookup" && expr$name == "id"
 }
 
+is_logical <- function(expr) {
+  truthy <- c(quote(TRUE), quote(FALSE), quote(true), quote(false))
+  bool <- vlapply(truthy, function(x) x == expr)
+  sum(bool) == 1
```
richfitz, line 290:
```suggestion
  truthy <- list(TRUE, FALSE, quote(true), quote(false))
  any(vlapply(truthy, identical))
```

I feel like use of `==` here might do coersions. You could also do this as:

```
(is.logical(expr) && !is.na(expr)) || (is.symbol(expr) && as.character(expr) %in% c("true", "false"))
```

richfitz (approved):
thanks - no doubt actual use will throw up more corner cases, but this this looks much less chill than before. Optional comments included

PR #30, opened 2023-07-11
New function for reading metadata

richfitz:
really simple function which does not yet exist!


R/outpack_metadata.R
```diff
@@ -1,3 +1,27 @@
+##' Read metadata for a particular id. You may want to use
```
r-ash, line 1: Could let users pass a query here too? Could just leave that for later and see if anyone asks for it, I suppose end users might not run this function very often

r-ash (approved):

R/outpack_metadata.R
richfitz: yeah, I wondered about this - but a "query" here is pretty complicated because the interface includes a bunch of things (e.g., parameters, search options, etc)

PR #32, opened 2023-07-12
Better parameter errors

richfitz:
This PR tidies up error messages for parameters that are not correct when running a report:

* if no `orderly::orderly_parameters()` call we include a suggestion that the report might want that
* if a parameter is missing but there are near misses amongst unused parameters, we suggest these
* if a parameter is unused and there are also optional parameters that are not given and near misses we suggest those.

In order to make the error messages easy to work with I've used `cli`'s error formatting, based on `rlang`: see https://www.tidyverse.org/blog/2021/12/rlang-1-0-0-errors/ though I think there might be other posts on this too.

The easiest way to see the error text is to try the examples changed in the tests, for example:

```
> check_parameters(list(thing = 1), list(things = NULL))
Error in `check_parameters()`:
! Missing parameters: 'things'
ℹ You have extra parameters, possibly misspelt?
• 'things': could be your 'thing'
Run `rlang::last_trace()` to see where the error occurred.
```

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -347,7 +347,7 @@ near_match <- function(x, possibilities, threshold = 2, max_matches = 5) {
   }
   d <- set_names(drop(adist(x, possibilities, ignore.case = TRUE)),
                  possibilities)
-  utils::head(names(sort(d[d < threshold], decreasing = TRUE)), max_matches)
+  utils::head(names(sort(d[d <= threshold])), max_matches)
 }
 
 
--- a/R/util.R
+++ b/R/util.R
@@ -339,3 +339,37 @@ string_starts_with <- function(sub, str) {
 string_drop_prefix <- function(sub, str) {
   substr(str, nchar(sub) + 1, nchar(str))
 }
+
+
+near_match <- function(x, possibilities, threshold = 2, max_matches = 5) {
+  if (length(possibilities) == 0) {
+    return(character())
+  }
+  d <- set_names(drop(adist(x, possibilities, ignore.case = TRUE)),
+                 possibilities)
+  utils::head(names(sort(d[d < threshold], decreasing = TRUE)), max_matches)
+}
+
+
+near_matches <- function(x, ...) {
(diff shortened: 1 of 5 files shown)


weshinsley (approved):
Looks nice!

r-ash (commented):
I like how this prints the ability to add info and having the colour highlighting is really great. What annoys me about these is that it's hard from that error to know what line has failed. 

Even with the backtrace 

```
> rlang::last_trace()
<error/rlang_error>
Error in `check_parameters()` at orderly2/R/run.R:114:2:
! Parameters given, but none declared
ℹ Did you forget 'orderly2::orderly_parameter()
---
Backtrace:
    ▆
 1. └─orderly2::orderly_run(...)
 2.   └─orderly2:::check_parameters(parameters, dat$parameters) at orderly2/R/run.R:114:2
Run rlang::last_trace(drop = FALSE) to see 2 hidden frames.
```

Say I had some loops over a set of iso3 codes and I want to run a report for all of them, because it doesn't show you the signature of the function call I don't know which one failed. I don't know why they think the arguments aren't an important part of the function call? It's fine if I only call a function once during whatever script I am running but we don't always know that. I can't see a way to make the arguments show up in the trace? 

I guess all depends on the user who will see these errors and maybe this is enough information people using orderly.

R/util.R
```diff
@@ -339,3 +339,37 @@ string_starts_with <- function(sub, str) {
...
 }
+
+
+near_match <- function(x, possibilities, threshold = 2, max_matches = 5) {
+  if (length(possibilities) == 0) {
+    return(character())
+  }
+  d <- set_names(drop(adist(x, possibilities, ignore.case = TRUE)),
+                 possibilities)
+  utils::head(names(sort(d[d < threshold], decreasing = TRUE)), max_matches)
```
r-ash, line 350:
```suggestion
  utils::head(names(sort(d[d <= threshold])), max_matches)
```

r-ash (approved):

PR #33, opened 2023-07-12
Add some docs

richfitz:
This vignette will never be finished, but it's ready to at least spin off other tickets from. I suspect there are plenty of half finished sentences and missing links to resolve over time (or here depending on severity!)

This PR adds some key orderly2 vignettes:

* orderly2: finally some background on what is orderly
* introduction: a how-to guide, similar to the intro orderly vignette
* migrating: some notes on migration; these won't survive contact with reality I expect

I've also copied in the two vignettes from outpack (querying and metadata) but have not yet worked on the plugins one.

When reviewing, can we distinguish between things that should be fixed here: wanton untruths, trailing sentences, things that are very confusing vs more that could be added - we'll get a decent list of that and make tickets.


vignettes/introduction.Rmd
```diff
@@ -0,0 +1,412 @@
...
+```r
+orderly2::orderly_strict_mode()
+```
+
+anywhere within your `orderly.R` script (conventionally at the top). We may make this more granular in future, but by adding this we:
+
+* only copy files from the source directory (`src/<name>/`) to the draft directory where the report runs (`draft/<name>/<packet-id>`) that were declared with `orderly2::orderly_resource`; this leaves behind any extra files left over in development
+* warn at the end of running a packet if any files are found that are not part of an artefact
+
+Using strict mode also helps `orderly2` clean up the `src/<name>` directory more effectively after interactive development (see next section.
```
kgaythorpe, line 349: missing right bracket

vignettes/metadata.Rmd
```diff
@@ -0,0 +1,218 @@
...
+
+Making the packet available allows it to be used as a dependency, allows serving that packet if you are acting as a location (over the http or file protocols), and guarantees that the files are actually present locally.
+
+## Adding metadata
+
+You can simply copy metadata as the file `.outpack/metadata/<packet id>` if it does not yet exist. This does not make it available to anything yet as it is not known from any location. Dangling metadata (that is, metadata present in this directory but not known anywhere) is currently mostly ignored.
+
+## Adding files
+
+If the repository uses a file store, you should fill this first, because it is much easier to think about. You can easily get the difference between files used by a packet (the list of files in the packet manifest) and what you already have in the file store by looking up each hash in turn.  You should then request and missing files and insert them into the store. This may leave "dangling" files for a while (files referred to by no packet) but that is not  problem.
```
kgaythorpe, line 126: that is not "a" problem.

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+
+## Structure of queries
+
+The most simple query is
+
+```r
+latest()
+```
+
+which finds the most recent packet; this is unlikely to be very useful without scoping - see below.
```
weshinsley, line 30: I've ended up a little confused in this doc - so returning to the top after reading it...

1. I think I'd find it easier if we started with a "useful" query. This one, if I understand it right, has us depending on the latest report in the archive, regardless of what the report was - and I agree that's unlikely to be useful without scoping (it probably doesn't make sense really)

2. So adding the scope is familiar if you're used to orderly-1, but then we have a "Warning we may remove this" at the top of the Scoping section... 

So I think maybe an approach is to decide whether or not we want scoping to be kept. If it does, then describe it up here as a helping hand for orderly-1 veterans, and say the new `name` parameter is the new and shiny way of doing it. Or otherwise, kill scoping off properly and don't reference it here; just go straight in with a useful-ish (and maybe more concrete/eg covid) example that uses both latest, and a named report.

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+knitr::include_graphics("images/dependency_terminology.png")
+```
+
+We could equivalently say
+
+* A is a dependency of B, B depends on A
+* A is used by B , B uses A
+* A is a parent of B, B is a child of A
+
+With the tree of dependencies built by outpack we might want to search for packets which have been used by another packet. We can use the query function `usedby(id)` to list all packets which are used by `id`. This will search recursively through all packets used by `id` and its parents and its parents parents and so on.
```
weshinsley, line 112:
```suggestion
With the tree of dependencies built by outpack we might want to search for packets which have been used by another packet. We can use the query function `usedby(id)` to list all packets which are used by `id`. This will search recursively through all packets used by `id` and its parents and its parents' parents and so on.
```

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+
+The optional second arg `immedate` is `FALSE` by default, if set to `TRUE` then we search only for immediate (e.g. level 1) dependencies.
```
weshinsley, line 114:
```suggestion
The optional second arg `immediate` is `FALSE` by default, if set to `TRUE` then we search only for immediate (e.g. level 1) dependencies.
```

weshinsley: This is for the package rather than the vignette really, but I wonder if `recurse={default TRUE, FALSE}` might be more intuitive than `immediate={default FALSE,TRUE}`

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+
+`usedby` can be simplified by using subqueries. Subqueries are denoted by curly braces `{}` and can either be named and passed in `subquery` arg or can be anonymous. The query below is equivalent to the above but uses a subquery for `C`.
+
+```r
+outpack_search(quote(usedby({C}) && parameter:year == 2022),
+               name = "A",
+               subquery = list(C = quote(latest(name == "C"))))
+```
+
+There are two important things to note about `usedby`:
```
weshinsley, line 151: I *think* you might need a newline before the `*` - this hasn't rendered as a bullet for me with devtools::build_vignettes()

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+
+We can combine `usedby` and `uses` in more complex searches, such as to find `D` from `C` we could run
+
+```
+outpack_query(
+  quote(usedby(single(uses(single(usedby(latest(name == "C")) && name == "A"))) && name == "E"))),
+  name = "D",
+  root = root)
+```
+
```
weshinsley, line 197: I almost feel there should be some practice exercises.

Also - are circular dependencies worth mentioning - presumably this isn't allowed?

weshinsley (changes requested):
Good stuff - it's going to be exciting to use all this. A bunch of comments / a few typos / queries on the way...

vignettes/orderly2.Rmd
```diff
@@ -0,0 +1,81 @@
...
+* by tracking the way data flows through an analysis, it allows the minimum amount of recalculation to be done on change, with only downstream parts triggered
+* with a shared workspace or online runner, allows a degree of collaboration so long as everyone is happy to be working with a constantly changing set of code and analysis artefacts
+
+We have designed `orderly2` for working patterns that do not suit the above. Some motivating reasons include:
+
+* Some nodes in the computational graph are very expensive to compute or require exotic hardware.
+* Workflows where the upstream data never settle, but for which we need to know which version of the data end up used in a particular analysis
+* Workflows where upstream analyses are used in many downstream analyses, and which the upstream developer may not know much about the downstream use
+* Nondeterministic analyses, e.g., those involving stochastic simulations, where rerunning a node is not expected to return the same exact numerical results, and so we can't rely on different users recovering the same results on different occasions.
```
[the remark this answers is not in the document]
richfitz: I'd be more concerned about that if the people levelling the original criticisms had any idea what they were talking about (and I really don't think many of the people you are talking about there have strongly held beliefs about reproducibility, though I'd not be so cynical as to suggesting some of those arguments were in bad faith). I certainly don't want to encourage people to set seeds without knowing *exactly* what they're doing as it's easy to get deterministic yet meaningless results that way...

vignettes/query.Rmd
```diff
@@ -0,0 +1,212 @@
...
+parameter:x > 1
+```
+
+Every "test" uses a boolean operator (`<`, `>`, `<=`, `>=`, `==`, or `!=`) and the left and right hand side can be one of:
+
+* a lookup into the outpack metadata (`parameter:x` is the value of a parameter called `x`, `name` is the name of the packet, and `id` is the id of a packet)
+* a lookup into the provided data `pars` (`this:x` is the value of `pars$x`)
+* a literal value (e.g., `"some_name"`, `1`, or `TRUE`)
+
+Tests can be grouped together `(`, `!`, `&&`, and `||` as you might expect:
```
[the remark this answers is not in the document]
richfitz: technically, `(` is a function but `)` is syntactic sugar (e.g. `as.list(quote((a + b)))`)

vignettes/migrating.Rmd
```diff
@@ -0,0 +1,99 @@
...
+* any R script can be the basis of an orderly report, and you can add orderly functions annotating inputs and outputs progressively
+
+## Database support has been moved into a plugin
+
+In version 1, we had built-in support for accessing data from SQL databases; this has moved within the [`orderly.db`](https://github.com/mrc-ide/orderly.db) plugin. All major features are supported.
+
+## No more commit
+
+## No more testing or development mode
+
```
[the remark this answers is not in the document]
richfitz: done

vignettes/introduction.Rmd
```diff
@@ -0,0 +1,412 @@
...
+```
+
+To run the report and create a new **packet**, use `orderly2::orderly_run()`:
+
+```{r}
+id <- orderly2::orderly_run("data", root = path)
+id
+```
+
+The `id` that is created is a new identifier for the packet that will be both unique among all packets (within reason) and chronologically sortable. A packet that has an that sorts after another was started before that packet.
```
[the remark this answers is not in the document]
richfitz: thanks

vignettes/query.Rmd
richfitz: Agreed, once the querying settles down a bit we should probably make this more accessible (volunteers welcome, but this will be a bit of a job I suspect, with lots of concrete examples)

vignettes/migrating.Rmd
```diff
@@ -0,0 +1,99 @@
...
+Next, migrate your source tree. This will be done *in place* so should be done on a fresh clone of your source git repository. For each report, we will examine your `orderly.yml` files and your script files (often `script.R`), delete these, and then write out a new `orderly.R` file that will adapt your report to work for `orderly2`.  It is possible that this will not be perfect and might need some minor tweaking but hopefully it will be reasonable. One thing that is not preserved (and we probably cannot do so) is the comments from the `yaml` but as these often refer to `yaml` formatting or `orderly1` features hopefully this is not too much of a problem.  You will probably want to manually tweak the generated code anyway, to take advantage of some of the new `orderly2` features such as being able to compute dependencies.
+
+If you are using OrderlyWeb, you probably need to pause before migrating, as the replacement is not yet ready.
+
```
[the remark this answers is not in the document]
richfitz: done

vignettes/introduction.Rmd
```diff
@@ -0,0 +1,412 @@
...
+write.csv(data.frame(x = 1:10, y = 1:10 + rnorm(10)), 
+          file.path(path, "src", "data", "data.csv"),
+          row.names = FALSE)
+writeLines(c(
+  'd <- read.csv("data.csv")',
+  "d$z <- resid(lm(y ~ x, d))",
+  'saveRDS(d, "data.rds")'),
+  file.path(path, "src", "data", "orderly.R"))
+```
+
```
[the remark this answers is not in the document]
richfitz: I've changed this now, not sure it's much better but we can always change it again later

vignettes/migrating.Rmd
```diff
@@ -0,0 +1,110 @@
...
+`orderly2` no longer requires a separate `orderly_commit()` call after `orderly_run()`; we no longer make a distinction between local draft and archive packets. Instead, we have added finer-grained control over where dependencies are resolved from (locally, or from some subset of your servers), which generalises the way that draft/archive was used in practice. See the "Dependency resolution" section of `?outpack_packet_use_dependency` (itself used from `orderly2::orderly_dependency`) for more details.
+
+## No more testing or development mode
+
+We have had two different, but unsatisfactory, mechanisms for developing an orderly report:
+
+* `orderly_test_start` (up to version 1.1.0)
+* `orderly_develop_start` (from 1.1.0 onwards)
+
+These worked by doing the initial setup, and copying of dependencies etc to a location you could work (a new draft for `orderly_test_start` and the source directoryt for `orderly_develop_start`). From `orderly2` you can just work directly within the source directory, and so long as your working directory is set to `src/<report-name>`, all `orderly2` commands will work as expected.
```
kgaythorpe, line 79: do people need to be extra careful on not committing outputs (orderly_develop_clean was great). also typo in directory

vignettes/migrating.Rmd
```diff
@@ -0,0 +1,110 @@
...
+
+* the changelog feature (we will implement something to support this)
+* extraction of secrets from vault (we will support this)
+* automatic handling of README files (we may implement support for this)
+
+# How to migrate
+
+There are two parts to a migration: updating the canonical copy of your orderly archive (ideally you only have one of these) and updating your source tree. These steps should be done via the `orderly2outpack` package.
+
+You should migrate your archive first. Do this for every archive that you want to retain (you might have archives stored lcoally, on production servers and on staging servers). Archive migration happens *out of place*; that is, we do not modify anything in the original location. If your archive is old and has been used with very old versions of `orderly1` it is possible that this process will have a few hiccups. Please let us know if that is the case.  The result of this process is that you will end up with a new directory that contains a new archive conforming to the `outpack` spec and containing `orderly2` metadata.
```
kgaythorpe, line 97: typo in locally

vignettes/orderly2.Rmd
```diff
@@ -0,0 +1,81 @@
...
+* it allows high-level parallelism by making obvious parts of the workflow that can be run concurrently
+* by tracking the way data flows through an analysis, it allows the minimum amount of recalculation to be done on change, with only downstream parts triggered
+* with a shared workspace or online runner, allows a degree of collaboration so long as everyone is happy to be working with a constantly changing set of code and analysis artefacts
+
+We have designed `orderly2` for working patterns that do not suit the above. Some motivating reasons include:
+
+* Some nodes in the computational graph are very expensive to compute or require exotic hardware.
+* Workflows where the upstream data never settle, but we need to know which version of the data ends up used in a particular analysis
+* Workflows where upstream analyses are used in many downstream analyses, and where the upstream developer may not know much about the downstream use
+* Nondeterministic analyses, e.g., those involving stochastic simulations, where rerunning a node is not expected to return the exact same numerical results, and so we can't rely on different users recovering the same results on different occasions.
```
kgaythorpe, line 63: is it worth noting that it is possible to set the seed ie. completely recover the results?

vignettes/query.Rmd
```diff
@@ -0,0 +1,213 @@
...
+  %\VignetteIndexEntry{Outpack query DSL}
+  %\VignetteEngine{knitr::rmarkdown}
+  %\VignetteEncoding{UTF-8}
+---
+
+Outpack includes a query DSL (domain specific language), extending the one used by orderly (see [`orderly1::orderly_search()`](https://www.vaccineimpact.org/orderly/reference/orderly_search.html)).
+
+Queries are used in identifying ids to pull in as dependencies, so rather than providing an identifier, you might want to depend on
+
+* the most recent copy of a packet with a particular name
```
kgaythorpe, line 14: can you add an example

kgaythorpe: with more than one option

kgaythorpe: ok I just read Wes' and your conversation- skip mine

vignettes/query.Rmd
```diff
@@ -0,0 +1,213 @@
...
+* The query inside `usedby` must return a single result. To ensure this it must either be a literal `id`, a call to `latest` or a call to `single`
+
+As well as searching up the dependency tree using `usedby` we can search down with the `uses` function. In the same setup above with reports `A`, `B` and `C` if we want to know the `id` of `C` which uses `A` we can find this by using
+
+```r
+outpack_search(quote(uses(latest(name == "A"))),
+               name = "C")
+```
+
+`uses` and `useby` can be combined to search more complex arrangements of dependencies. If we have something like
```
kgaythorpe, line 163: should this be usedby?

vignettes/orderly2.Rmd
```diff
@@ -0,0 +1,85 @@
...
+---
+
+`orderly2` is a package designed with two complementary goals in mind:
+
+1. to make analyses reproducible without significant effort on the part of the analyst
+2. to make it easy to collaborate on analyses by allowing easy sharing of artefacts from an analysis amongst a group of analysts.
+
+In this vignette we will expand on these two aims, and show that the first one is a prerequisite for the second. The second is more interesting though and we start there. If you just want to get started using `orderly2`, you might prefer `vignette("introduction")` and if you are already familiar with version 1 you might prefer `vignette("migrating")`.
+
+# Collaborative analysis
```
r-ash, line 17: This section is great!

vignettes/introduction.Rmd
```diff
@@ -0,0 +1,412 @@
...
+This creates a report that has a single parameter `n_samples` with a default value of 10. We could have used
+
+```r
+orderly2::orderly_parameters(n_samples = NULL)`
+```
+
+to define a parameter with no default, or defined multiple parameters with 
+
+```r
+orderly2::orderly_parameters(n_samples = 10, distribution = "normal")`
```
r-ash, line 223:
```suggestion
orderly2::orderly_parameters(n_samples = 10, distribution = "normal")
```

vignettes/migrating.Rmd
```diff
@@ -0,0 +1,112 @@
...
+```
+
+would end up within an `orderly.R` script that looks like:
+
+```r
+orderly2::orderly_parameters(n_min = 10)
+orderly2::orderly_dependency("raw_data", "latest", 
+                             use = c("raw_data.csv" = "data.csv"))
+orderly2::orderly_resource("metadata.csv")
+orderly2::orderly_artefact("Processed data", "data.rds"))
```
r-ash, line 49:
```suggestion
orderly2::orderly_artefact("Processed data", "data.rds")
```

r-ash (changes requested):
Couple of small changes requested but this looks great.

For next vignette would be useful to see some details about how locations & remotes work in orderly2

PR #34, opened 2023-07-13
Allow environment lookup

richfitz:
This PR adds something required for us to be able to practically have a loop where we do:

```
for (r in regions) {
  orderly2::orderly_dependency("data", "parameter:region == environment:region", ...)
}
```

(there's still some issues around how we do the renaming of files in the dots there, ignored for now). This PR adds another explicit lookup.

Two things to be aware of here:

* anything that uses `parameter:x` could get away with `environment:x` but probably people should prefer the former, and we might prevent the latter in strict mode. Thoughts on this very welcome (including if they are an issue for now or later)
* is the name `environment:` right? I wondered about `local:` or `dynamic:` - this will form part of the spec so a general term is useful, but it will also be almost impossible for the rust version (for example) to support it

This PR is related to #40, which also uses the environment

### DIFF
--- a/R/metadata.R
+++ b/R/metadata.R
@@ -262,13 +262,17 @@ orderly_dependency <- function(name, query, use) {
   if (ctx$is_active) {
     outpack_packet_use_dependency(ctx$packet, query, use,
                                   search_options = search_options,
+                                  envir = ctx$env,
                                   overwrite = TRUE)
   } else {
     id <- outpack_search(query, parameters = ctx$parameters,
-                         options = search_options, root = ctx$root)
+                         envir = ctx$env,
+                         options = search_options,
+                         root = ctx$root)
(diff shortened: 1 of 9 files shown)


r-ash (commented):
This kind of looks like a more general version of `this:`?

Can you remind me where we use that, when running 

```
orderly::orderly_search("parameter:x == this:x", parameters = list(x = "value"))
```

Could we reuse that or combine these instead of adding a separate thing?

```
for (r in regions) {
  orderly::orderly_dependency("data", "parameter:region == this:r")
}
```

Maybe there more different than I am realising

r-ash (approved):
Ok, chatted about the difference between `this` and `environment` we want to keep `this` as special as it will always refer to the parameters from the currently running packet, where `environment` could be anything from the environment.

PR #36, opened 2023-07-14
Overwrite on copy by default

richfitz:
This PR changes the default behaviour of using dependencies and global resources to copy with `overwrite = TRUE` on; this is on by default with outpack_ functions and on without the ability to change in `orderly_` functions as these are meant to be less tuneable.

This matches the behaviour of orderly1, and is required to make the new approach to development mode work, as we'd expect people to copy files in a few times while developing.


weshinsley (approved):
Looks good to me!

PR #37, opened 2023-07-17
Allow empty queries to return everything

richfitz:
This is required to allow:

```
orderly2::orderly_search(name = "foo")
```

to do the same thing as:

```
orderly2::orderly_search(quote(name == "foo"))
```

which feels desirable (it certainly did while writing the docs!)

### DIFF
--- a/R/query.R
+++ b/R/query.R
@@ -3,7 +3,7 @@
 ##'
 ##' @title Construct outpack query
 ##'
-##' @param expr The query expression
+##' @param expr The query expression. A `NULL` expression matches everything.
 ##'
 ##' @param name Optionally, the name of the packet to scope the query on. This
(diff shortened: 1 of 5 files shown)


R/query.R
```diff
@@ -71,6 +74,7 @@ query_parse <- function(expr, context, subquery_env) {
       }
       expr <- expr[[1L]]
     }
+  } else if (is.null(expr)) {
```
weshinsley, line 77:
```suggestion
  } else if (is.null(expr)) {
    # Nothing needed for null expression
```

weshinsley: Just wondering if it'll make line 77 look more "intentional" when we look at it later...

weshinsley (approved):

weshinsley (approved):

PR #38, opened 2023-07-17
Fix logging schema

richfitz:
This PR updates the incorrect configuration schema for logging:

* use `threshold`, not `log_level` (the former is the level we filter logs down to, the latter is the level they are emitted with), and correct the type (was integer, should be an enum)
* make `threshold` and `console` required properties if either a `logging` element is present

### DIFF
--- a/inst/outpack/schema/config.json
+++ b/inst/outpack/schema/config.json
@@ -31,13 +31,14 @@
         "logging": {
             "type": "object",
             "properties": {
-                "log_level": {
-                    "type": "integer"
+                "threshold": {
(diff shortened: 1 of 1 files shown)


hillalex (commented):
Is this really optional? I was getting errors running reports when the `logging` block was missing from my config.

richfitz:
Let me check - I wasn't sure what was triggering your original error from the ticket

richfitz:
Right, I think I see what was probably your problem; let me know if that fixes it

hillalex (approved):
yep, perfect, thanks!

PR #44, opened 2023-07-20
Clean up generated files from source directory 

richfitz:
This PR goes some way to support the old `orderly_develop_status` and `orderly_develop_clean` workflow from before. There are differences though as we no longer have any expectation of a full file list from reading the orderly metadata.

The approach here then is driven from the files we see, which we then try and classify  - we can score things as the orderly.R file, resources, global resources, dependencies, or artefacts. The last three types need deleting when we clean up.

@kgaythorpe made the excellent suggestion of encouraging people to use .gitignore, and I remembered that libgit2 has the ability to tell if a file *is* gitignored, so with very recent gert we can use this to find files that the user probably wants gone: https://github.com/r-lib/gert/pull/204

There are some fixes here to metadata handling that were required to make this work.

I'd like some feedback on if the names are right, but at the same time we have a big rename coming up soon. I've indicated in the doc that the status type is unstable, andI imagine that we'll put a format method on that at some point


r-ash (approved):
This looks great to me, using gitignore is an excellent idea

PR #47, opened 2023-07-24
Import and rewrite the plugin vignette

richfitz:
Minor interface changes here, just enforcing consistency between the schema and serialise arg and requiring that if the plugin returns data that we know what do with it.

Certainly this will get changed again once we start on rethinking how configuration works, but this at least is consistent with orderly.db


vignettes/plugins.Rmd
```diff
@@ -0,0 +1,377 @@
...
+
+To use a plugin for an individual report, functions from the plugin should be used, which configure and use the plugin.
+
+Finally, we can save information back into the final `orderly2` metadata about what the plugin did.
+
+With the yaml-less design of `orderly2` (see `vignette("migrating")` if you are familar with `orderly1`), the line between a plugin and just package code is fairly blurred, but reasons for writing a plugin are typically that you want to make something easier in reports, and you want that action reflected in the orderly metadata.
+
+## An example
+
+As an example, we'll implement a stripped down version of the database plugin that inspired this work (see [`orderly.db](https://github.com/mrc-ide/orderly.db) for a fuller implementation). To make this work we need functions:
```
r-ash, line 42:
```suggestion
As an example, we'll implement a stripped down version of the database plugin that inspired this work (see [`orderly.db`](https://github.com/mrc-ide/orderly.db) for a fuller implementation). To make this work we need functions:
```

r-ash (approved):
This looks great, one tiny typo I noticed. I'll give this a go at some point and try and write a plugin for pulling input data from sharepoint and can look out for any confusing bits then. But this looks great on just reading through.

PR #48, opened 2023-07-24
Drop configurability of global resources

richfitz:
Nobody (except one case) actually modified this. We can always add it back in later.

With this out, only plugins use the yml, and soon we can squash them out.

I do wonder (somewhat separately) if we should move from `global` to `shared`?


r-ash (approved):
Looks good - I like `shared` too

PR #49, opened 2023-07-24
Allow orderly_copy_files to accept a query

richfitz:
~Merge after #46, contains those commits~

This PR updates orderly_copy_files to accept a query, using the same approach as orderly_metadata_extract (#39). There was already quite a bit of overlap between the arguments as passed through to the original function and to `orderly_search` so this feels fairly nice, even though all the additional arguments must now be named.


r-ash (approved):
This looks good, I like the new interface

PR #51, opened 2023-07-26
mrc-4416: harmonise orderly and outpack root and init

richfitz:
This was much grosser than expected, but the end result is nice and consistent, and I've updated the docs and error messages everywhere.

Now there is only a single type of root, and it might have orderly data too. Functions that need a root can indicate if they need the orderly bits.

Initialising a partially initialised root is now easy in both directions - this will happen when "upgrading" an outpack one to use orderly (rare) or when adding outpack to a cloned source tree (common). The errors describe what need doing here too.

A huge amount of churn, some of which is done by just not hitting the root at all and using outpack_metadata


R/root.R
```diff
@@ -1,87 +1,233 @@
...
+  'minimum_orderly_version: "1.99.0"'
+}
+
+
+## There's quite a few scenarios here:
 ##
-## Things that will probably come in without too much change:
+## * find an existing outpack root that is not an orderly root
+##   - sometimes require that we upgrade (e.g., run) [require orderly t]
+##   - sometimes just work with it (e.g., search, extract, copy) [r o f]
```
r-ash, line 121: What's `require orderly t`? And `r o f`?

r-ash (approved):
This looks great, one question about some cryptic looking docs but changes look good

R/root.R
richfitz: wtf

richfitz: so, I think this was "require orderly = TRUE" and "require orderly = FALSE"

PR #59, opened 2023-07-27
Better deserialisation of custom metadata

richfitz:
This PR does a few things that have annoyed me:

* the orderly metadata is now nicely deserialised on metadata load, which makes it much nicer to work with
* the weird wrangling of the schema to have custom-metadata-in-custom-metadata is gone, plugins now sit at the top-level outpack custom level metadata
* some tweaks to outpack_metadata_extract that follow from that to allow extraction of paths that have dots in them


r-ash (approved):
Looks good!

PR #61, opened 2023-07-28
mrc-4414: curate gitignore files automatically

richfitz:
Some work still to go, particularly with programmatic deps etc, but this is probably already useful.

Not sold on `(root)` for the root _at all_ and I think that detecting the directory would be better here, but we need to do that in a few places and don't yet have the support do so generally (mrc-4448)

Also not done is building these automatically, e.g., after running orderly_init, or even after a report is run

As discussed, this no longer prompts the user before making changes


r-ash (approved):
Looks great

PR #65, opened 2023-08-01
Start on collaboration/locations vignette

richfitz:
Lots to do here still, but it gets a few ideas down at least

### DIFF
--- a/vignettes/collaboration.Rmd
+++ b/vignettes/collaboration.Rmd
@@ -132,7 +132,7 @@ Suppose we have two researchers, [Alice and Bob](https://en.wikipedia.org/wiki/A
 
 Both Alice and Bob clone this repo onto their machines, using their favourite git client (at the terminal with `git`, via GitHub desktop, or as here with [`gert`](https://r-lib.github.io/gert))
 
-At the moment, including hidden files (except `.git`, Alice sees:
+At the moment, including hidden files (except `.git`), Alice sees:
 
 ```{r, as = "alice", inwd = path_alice}
 fs::dir_tree(".", all = TRUE, glob = ".git", invert = TRUE)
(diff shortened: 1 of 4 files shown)


vignettes/collaboration.Rmd
```diff
@@ -0,0 +1,261 @@
...
+# An example
+
+Suppose we have two researchers, [Alice and Bob](https://en.wikipedia.org/wiki/Alice_and_Bob), who are collaborating on a piece of analysis. For the purposes of this example (and we find the most common situation in practice) they share a source tree with their analysis in git.
+<!-- The source for the repository that they are using is on GitHub as
+[`reside-ic/orderly-example`](https://github.com/reside-ic/orderly-example).
+-->
+
+Both Alice and Bob clone this repo onto their machines, using their favourite git client (at the terminal with `git`, via GitHub desktop, or as here with [`gert`](https://r-lib.github.io/gert))
+
+At the moment, including hidden files (except `.git`, Alice sees:
```
r-ash, line 135:
```suggestion
At the moment, including hidden files (except `.git`), Alice sees:
```

vignettes/collaboration.Rmd
```diff
@@ -0,0 +1,261 @@
...
+
+**Central server, final copy is run here**: In this case, Alice and Bob can both run things on the server, and these are the "canonical" copies, nothing is pushed to the server.  Reports can be run on the server by `ssh`-ing onto the server and running packets manually (TODO: mrc-4412) or using the web interface (this was possible with `OrderlyWeb` and will be possible again with `packit`).
+
+This approach works well where a primary goal is confidence that the packets everyone works with as dependencies are created under a "clean" environment with no unexpected global dependencies. With the web version, you can also enforce things like only running off the default branch.  Alice and Bob will thenend up with a collection of packets in their local archives that are a mix of canonical ones (from the server) and locally-created ones, but the local ones are always a dead end as they never get shared with anyone. As a result, Alice and Bob may delete their archive directories without any great concern.
```
r-ash, line 217:
```suggestion
This approach works well where a primary goal is confidence that the packets everyone works with as dependencies are created under a "clean" environment with no unexpected global dependencies. With the web version, you can also enforce things like only running off the default branch.  Alice and Bob will then end up with a collection of packets in their local archives that are a mix of canonical ones (from the server) and locally-created ones, but the local ones are always a dead end as they never get shared with anyone. As a result, Alice and Bob may delete their archive directories without any great concern.
```

vignettes/collaboration.Rmd
```diff
@@ -0,0 +1,261 @@
...
+# Interaction between source tree and `.outpack`
+
+It is important that the `.outpack` directory is *not* shared via git; we warn about this now, and you can use `orderly2::ordery_gitignore_update()` to automatically create a suitable `.gitignore` file that will prevent it being accidentically committed.
+
+However, 
+
+If Alice and Bob were starting on a new machine they would:
+
+* clone their source repository
+* run `orderly::orderly_init()` from within the directory
```
r-ash, line 261: This ends fairly abruptly? Was there a final point to make? If they run `orderly::orderly_init()` from within the directory it will create the `.outpack` directory?

vignettes/collaboration.Rmd
```diff
@@ -0,0 +1,261 @@
...
+
+The location system is somewhat extensible, but the details of this are subject to change.  There is [`outpack.sharepoint`](https://github.com/mrc-ide/outpack.sharepoint) for using SharePoint, and we may write other backends later.
+
+All the documentation below will behave the same regardless of where the location is stored and the mechanism of transport, so we can focus instead on workflows.
+
+# An example
+
+Suppose we have two researchers, [Alice and Bob](https://en.wikipedia.org/wiki/Alice_and_Bob), who are collaborating on a piece of analysis. For the purposes of this example (and we find the most common situation in practice) they share a source tree with their analysis in git.
+<!-- The source for the repository that they are using is on GitHub as
+[`reside-ic/orderly-example`](https://github.com/reside-ic/orderly-example).
```
r-ash, line 130: You don't want to make this visible to the reader? I could see it being useful to be able to read the examples

r-ash (approved):
Looks great, one comment about the last point which feels like it trails off but otherwise just couple of tiny typos I spotted

vignettes/collaboration.Rmd
richfitz: this does not exist yet!

vignettes/collaboration.Rmd
richfitz: I've expanded this somewhat now

PR #66, opened 2023-08-01
Explain why queries match (or don't match)

richfitz:
This is the first PR in a series on explainable queries. This one lats the groundwork, by providing explanations for the simplest queries. There's a big comment in there on one of the next steps (explaining near misses in equality tests), and this also needs wiring into the error messages when dependency resolution fails.


R/query_explain.R
```diff
@@ -0,0 +1,118 @@
...
+  ## In the case where we have exactly one 'n == 0' case we can try
+  ## and find more near misses, but this will be a follow-on PR. The
+  ## idea is that we will identify the offending bit like:
+  ##
+  ## > parameter:x == value
+  ##
+  ## Then we evaluate the rest of the query - all the bits without
+  ## this chunk to get the context. Then look to see how we can
+  ## permute the name of the parameter to get a near miss, or the
+  ## value to get a near miss.
```
r-ash, line 51: Yeah this sounds great, I wonder if also as an extension could use string distance on potential packet names? To give some suggestions for `name == "y"` if `"y"` doesn't find any

r-ash (approved):
This looks like a great first cut, code is easy to follow. Think this will already be really useful as is as even if it doesn't break apart all the queries users can do that manually and call `orderly_query_explain`.

I wonder if worth adding the ability later to get some info about the packets too, we return e.g.

```
B (parameter:b == 2): 3 packets
```
I wonder if we should give users a way to print a summary of these 3 packets. That would also be generally useful I think. Though what forms a summary of a packet not obvious! id, name and parameter values? Wonder if you could do that with rlang stuff nicely somehow to not print the details immediately but show them after running a 2nd command maybe. I'd be up for trying to add that if it sounds like an idea.

R/query_explain.R
richfitz: yeah, I was expecting to reuse the "did you mean" bits that we have already

richfitz:
> I wonder if we should give users a way to print a summary of these 3 packets

Yeah, perhaps an option to the print method that shows more detail? Very open to suggestions on how we can do this in a way that is easy to work with. Note that the actual object returned should be fairly pleasant to work with too

PR #67, opened 2023-08-02
mrc-4463: drop location id, name becomes primary key

richfitz:
I was starting to move the schemas over, but realised that we never use this bit of indirection - and I don't even remember why it is here.

I think that the idea was that we could track back from a packet to the original location where it was generated (like a fingerprint) but that never made it through to the rest of the package. So now we just let the name be the primary key, which is what git does anyway.

From the tests you can see that the fallout is very minor - we really weren't using this "feature" at all for anything

The workflows should get the outpack_server branch pin updated before merge


hillalex (approved):
Nice, much simpler!

PR #68, opened 2023-08-02
Explicitly pass environment to context

richfitz:
This simplifies some weird code, and will be more reliable in practice I think.

I've tried to come up with a failing test, that does not work on main, but can't do so - it's easy to trigger the failure interactively though...

### DIFF
--- a/tests/testthat/test-run.R
+++ b/tests/testthat/test-run.R
@@ -138,7 +138,7 @@ test_that("can run orderly with parameters, without orderly", {
 })
 
 
-test_that("can ruin orderly with parameters, without orderly, globally", {
+test_that("can run orderly with parameters, without orderly, globally", {
   path <- test_prepare_orderly_example(c("parameters", "depends-query"))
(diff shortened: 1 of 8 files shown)


tests/testthat/test-run.R
```diff
@@ -138,6 +138,20 @@ test_that("can run orderly with parameters, without orderly", {
 })
 
 
+test_that("can ruin orderly with parameters, without orderly, globally", {
```
r-ash, line 141:
```suggestion
test_that("can run orderly with parameters, without orderly, globally", {
```

"can ruin orderly" :laughing:

r-ash: Actually what does "without orderly" mean here?

r-ash (changes requested):

tests/testthat/test-run.R
richfitz: means without orderly_run (i.e., interactively)

r-ash (approved):

PR #74, opened 2023-08-04
Re-enable windows test

richfitz:
Seeing if https://github.com/mrc-ide/outpack_server/pull/28 is enough to fix this


plietar:
Hadn't seen this and opened https://github.com/mrc-ide/orderly2/pull/170 instead. Seems to work now.

PR #75, opened 2023-08-04
Import schemas from outpack

richfitz:
This PR uses the schemas from outpack, with quite a bit of associated cleanup to harmonise use of schemas from orderly2, outpack and from plugins.

In this approach, the logging and orderly custom schemas are orderly2 things, everything else is outpack


r-ash (approved):
Looks good, will leave open to update that branch in the `update_schemas` script

PR #76, opened 2023-08-04
mrc-4467: start on dependencies docs

richfitz:
Incomplete, but better than nothing?

Lots of room to add more, but at some point someone else should have a go...


r-ash (approved):
This looks great

PR #80, opened 2023-08-08
Log dependency resolution, throwing appropriate errors

richfitz:
This PR does a few related things:

* Log to screen and log.json the resolution of dependencies (orderly1 did this and it was useful)
* Use the query explanation as part of the error message
* Fix a bug in deparsing a `NULL` query

### DIFF
--- a/vignettes/dependencies.Rmd
+++ b/vignettes/dependencies.Rmd
@@ -175,7 +175,7 @@ The error suggests running `rlang::last_error()$explanation` for more informatio
 rlang::last_error()$explanation
 ```
 
-This is an `orderly_query_explain` object, which tries to come up with reasons why your query might not have matched; we'll expand this in future so let us know what you might like to see.
+This is an `orderly_query_explain` object, which tries to come up with reasons why your query might not have matched; we'll expand this in the future so let us know what you might like to see.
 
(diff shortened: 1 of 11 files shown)


vignettes/dependencies.Rmd
```diff
@@ -159,6 +159,52 @@ Here the query `latest(parameter:cyl == this:cyl)` says "find the most recent pa
...
+
+The error message here tries to be fairly self explanatory; we have failed to find a packet that satisfies our query` latest(parameter:cyl == this:cyl && name == "data")`; note that the report name `data` has become part of this query, so there are two conditions being matched on.
+
+The error suggests running `rlang::last_error()$explanation` for more information, which we can do:
+
+```{r}
+rlang::last_error()$explanation
+```
+
+This is an `orderly_query_explain` object, which tries to come up with reasons why your query might not have matched; we'll expand this in future so let us know what you might like to see.
```
weshinsley, line 178:
```suggestion
This is an `orderly_query_explain` object, which tries to come up with reasons why your query might not have matched; we'll expand this in the future so let us know what you might like to see.
```

weshinsley (approved):
Very nice!
I found one typo...

PR #83, opened 2023-08-14
Tidy up packet pull

richfitz:
This is gross sorry, it uncovered more weird than I was expecting.

This PR tidies up the way that packets are pulled. It was motivated by two things:

1. when pulling 500MB of packets on Katy's machine, nothing happened for a while which was alarming, as no status update is printed
2. the algorithm for working out what to pull was packet-by-packet which was hard to reason about and not symmetric with the simpler pull algorithm

Unfortunately when I started tidying these things up, it just sort of kept growing :(

The easiest way to review is probably to largely ignore the red side and just look at the new implementation, which should be fairly easy to follow. This is the logic I expect we will implement in the Python (and possibly Rust) version

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -591,9 +591,9 @@ location_build_pull_plan_location <- function(packets, location, root, call) {
     allow_no_locations = length(packets$fetch) == 0)
   ## Things that are found in suitable location:
   candidates <- root$index$location(location_name)
-  msg <- setdiff(packets$fetch, candidates$packet)
-  if (length(msg) > 0) {
-    extra <- setdiff(msg, packets$requested)
+  missing <- setdiff(packets$fetch, candidates$packet)
+  if (length(missing) > 0) {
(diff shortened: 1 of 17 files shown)


R/location.R
```diff
@@ -696,3 +733,61 @@ mark_packets_orphaned <- function(location, packet_id, root) {
...
+## * copy from the file store into the final location
+##
+## So in the case where a hash is only present once in a chain of
+## packets being pulled this will be one too many hashes and one too
+## many copies.
+##
+## However, this approach makes the logic fairly easy to deal with,
+## and copes well with data races and corruption of data on disk
+## (e.g., users having edited files that we rely on, or editing them
+## after we hash them the first time).
```
r-ash, line 754: That sounds good!

r-ash (approved):
Trivial comment about some variable name but changes look great

PR #90, opened 2023-08-21
mrc-4548: nicer printing for cleanup status

richfitz:
This PR adds cli-based printing to the cleanup status, and adds a section on using it into the docs.

There's a sneaky feature addition of not telling the user to clean up a .gitignore, these are always fine


r-ash (approved):
Nice, this looks great

PR #93, opened 2023-08-25
Make greater/less than query tests not match with bool and character types

r-ash:
This was quite fiddly to get right but this will now
* Stop matching (from type coersion) if you use >, >=, <, <= and either side of the query uses a logical or character type


tests/testthat/test-query-search.R
```diff
@@ -960,3 +960,67 @@ test_that("can search for queries using boolean", {
...
+  x <- create_random_packet(root, "x", list(a = 2))
+  y <- create_random_packet(root, "y", list(a = "test"))
+
+  err <- expect_error(
+    orderly_search(quote(parameter:a > 1), root = root),
+    "Cannot use test '>' with supplied data type")
+  expect_equal(err$body,
+               c(x = "Search term 'parameter:a' of type 'character' is invalid",
+                 i = "This test can be used with types:",
+                 "*" = "numeric"))
```
r-ash, line 1025: See this part, if two different reports use the same parameter with different types this is going to error. This won't be super consistent say if they did `name == "x" && parameter:a > 1` I think this will be fine if it does that index filtering on name first.

At the least we probably need to put the offending packet ID in which is causing this error to be thrown?

But maybe we should be more permissive.. not sure

r-ash: Fixed this up now, we'll no longer error but instead not use type coersion in the matching. So for the example above report y won't match but x will. No error will be thrown.

richfitz (approved):
this feels very easy to reason about

PR #95, opened 2023-08-30
Return something useful from orderly_resource

richfitz:
As requested by @sangeetabhatia03, this PR makes the result of special in-report functions more useful:

* `orderly_resource`: returns the computed set of resources (accounting for any directory expansion)
* `orderly_shared_resource`: data.frame of mapping of old file to new file
* `orderly_dependency` - id, name and mapping of old to new file

@lmhaile is needing the mapping for `orderly_dependency`, so best get that in at the same time

I can't see much interesting to return from `orderly_strict_mode`, `orderly_description`, `orderly_parameters` or `orderly_artefact`


r-ash (approved):
This looks super handy

PR #97, opened 2023-09-01
Clean up orphans

richfitz:
Merge after #96, contains those commits

This PR adds support for cleaning up orphan metadata


R/prune.R
```diff
@@ -0,0 +1,52 @@
...
+  if (length(id) == 0) {
+    return(invisible(id))
+  }
+
+  idx <- new_query_index(root, orderly_search_options(location = local))
+  is_used <- lengths(lapply(id, idx$get_packet_uses, Inf)) > 0
+  if (any(is_used)) {
+    cli::cli_alert_info(
+      paste("Can't prune {sum(is_used)} orphan packet{?s}, as",
+            "{?it is/they are} referenced by other packets"))
```
r-ash, line 37: Worth telling them which `id` couldn't be pruned?

r-ash (approved):
Looks good

R/prune.R
richfitz: I wondered, but also I wonder if this could be potentially a very long list. Also it's not actionable - there's no way of dropping the metadata in this case

PR #103, opened 2023-09-08
Basic interface to extract graph metadata

richfitz:
This PR just puts some code in place to pull the bits that we might want to be able to create a DAG over orderly packets (**not** yet the source reports, which is quite a different problem, especially if people lean into the orderly2 features).

This is not yet exported to the users because we need a nice way to display it, but will hopefully be something we connect up to a nice js package

### DIFF
--- a/R/graph.R
+++ b/R/graph.R
@@ -8,16 +8,21 @@ orderly_graph_packets <- function(from = NULL, to = NULL,
 
   metadata <- root$index$data()$metadata
   if (!is.null(from)) {
-    dat <- graph_packets_from(from, metadata)
+    if (is.null(metadata[[from]])) {
+      cli::cli_abort("Packet '{from}' does not exist for 'from'")
+    }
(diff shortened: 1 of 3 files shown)


R/graph.R
```diff
@@ -0,0 +1,73 @@
...
+  } else {
+    dat <- graph_packets_to(to, metadata)
+  }
+
+  dat
+}
+
+
+graph_packets_to <- function(to, metadata) {
+  validate_outpack_id(to)
```
r-ash, line 21: Should we test that this `id` actually exists? Get a pretty ugly error if it metadata[[p]]$depends is `NULL`

```
Error in if (nrow(depends) > 0) { : argument is of length zero
```

richfitz: yeah, these are both gross. I've added fixes and tests for both, thanks

r-ash (approved):

PR #105, opened 2023-09-14
Add another test to solidify groupings precedence

r-ash:
Added this to conform to check some behaviour before implementing the equivalent in Rust

### DIFF
--- a/tests/testthat/test-query-search.R
+++ b/tests/testthat/test-query-search.R
@@ -961,6 +961,7 @@ test_that("can search for queries using boolean", {
     character(0))
 })
 
+
 test_that("query with invalid types does no type coersion", {
   root <- create_temporary_root(use_file_store = TRUE)
   x <- create_random_packet(root, "x", list(a = TRUE))
@@ -984,6 +985,7 @@ test_that("query with invalid types does no type coersion", {
     character(0))
 })
 
+
 test_that("query with mixed types returns results with valid comparison", {
   root <- create_temporary_root(use_file_store = TRUE)
   x <- create_random_packet(root, "x", list(a = 2))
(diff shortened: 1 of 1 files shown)


tests/testthat/test-query-search.R
```diff
@@ -993,3 +995,31 @@ test_that("query with mixed types returns results with valid comparison", {
     orderly_search(quote(parameter:a > 1), root = root),
     x)
 })
+
+
+test_that("&& takes precedence over ||", {
```
richfitz, line 1000: I am not sure that this is precendence but associativity - https://en.wikipedia.org/wiki/Operator_associativity Python does look to have the same associativity here which is good.

One thing that we could do is to spot things like A && B || C and require disambiguation (after the parse phase). So allow any number of the _same_ operator chained (`A || B || ... || Z` or `A && B && ... && Z`) but within a chain require that any switch between the two is grouped with parentheses. Some C linters will require this

richfitz (approved):
Happy to merge this, but also wonder if we should forbid it?

tests/testthat/test-query-search.R
r-ash: I think it is precedence no? From that article looks like if && and || were right-to-left associative we would have
`A && B || C` be interpreted as `A && (B || C)` but this is saying `&&` always takes precedence so it is `(A && B) || C` and in the case `A || B && C` it is `A || (B && C)`

Like this table for C https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#Operator_precedence

If we didn't allow it then it would mean it was not associative?

richfitz: You're right! See `?Syntax` for the full ordering. Looks like python does the same: https://www.programiz.com/python-programming/precedence-associativity

Issue #106, opened 2023-09-18
Orderly2 is sensitive to file extension case 


pwinskill:
This may be intentional, but it just tripped me up.
In R I can save and load a `.rds` object successfully when using a mixture of upper and lower case for the file extension. So the following works:
```
x <- 1:10
dir <- tempdir()
saveRDS(x, paste0(dir, "/x.RDS")) # Save with upper case
y <- readRDS(paste0(dir, "/x.rds")) # Load with lower case
```
I just ecountered an issue where an `orderly2::orderly_dependency` couldn't be found as I had a similar mis-match of upper and lower case file extensions. So in report1 I had:
``` 
orderly2::orderly_artefact(
  description = "c",
  files = "x.RDS"
)
```
and in report 2 I had:
```
orderly2::orderly_dependency(
  name = "report 1",
  query = "latest()",
 files = "x.rds"
)
```

richfitz:
This is intentional yes. It's a quirk of windows (and mac) that their filesystems are case insensitive; this causes all sorts of issues when people migrate things onto Linux systems and then things start failing. For example, your code fails for me on this machine:

```
> x <- 1:10
> dir <- tempdir()
> saveRDS(x, paste0(dir, "/x.RDS")) # Save with upper case
> y <- readRDS(paste0(dir, "/x.rds")) # Load with lower case
Error in gzfile(file, "rb") : cannot open the connection
In addition: Warning message:
In gzfile(file, "rb") :
  cannot open compressed file '/tmp/Rtmp9LZH88/x.rds', probable reason 'No such file or directory'
```

This portability issue becomes more pronounced when you start using the server (once it exists!) as that will run on Linux.

What orderly1 had, and what we'll put back at some point, is some detection that you had created a file with a non-canonical case. Although in your case above that would not actually have helped because you created a file `x.RDS` just like you said you would.

Probably we'll add something to help with near misses on filenames (we need this anyway) so this would have the `orderly2::orderly_dependency` call fail with a hint that it would succeed if you used `x.RDS` and not `x.rds`

PR #110, opened 2023-09-29
Check that files have the canonical case

richfitz:
~Merge after #109, as it contains that changeset.~

This PR does the harder half of the fix to #106, and checks that when files are referred to that they have the correct case. I've used `cli_abort` over `stop()` here so that's added to the churn. We should try and squash out more of the remaining `stop()` cases soon, and I'll follow this PR with one that sorts out all the assertion functions


r-ash (approved):
This looks good, I like giving the context in the error message about why we don't allow it

PR #113, opened 2023-10-11
Allow use of NULL for name in dependency

richfitz:
This allows use of an id as a query without a name, it would also allows for queries that might compute a name, I guess. Mostly for completeness, but I made the ticket so must have run into a situation where this might have been preferable?

### DIFF
--- a/R/metadata.R
+++ b/R/metadata.R
@@ -247,7 +247,9 @@ static_orderly_artefact <- function(args) {
 ##' @return Undefined
 ##' @export
 orderly_dependency <- function(name, query, files) {
-  assert_scalar_character(name, call = environment())
+  if (!is.null(name)) {
+    assert_scalar_character(name, call = environment())
+  }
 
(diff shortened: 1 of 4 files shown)


R/metadata.R
```diff
@@ -276,6 +278,9 @@ static_orderly_dependency <- function(args) {
   query <- args$query
   files <- args$files
 
+  static_name <- static_string(name)
+  has_name <- !is.null(static_name) || is.null(name)
```
weshinsley, line 282: Should there be `!` before `is.null(name)` ?

weshinsley (changes requested):
Not quite sure on the logic of the NULL name - but might be a typo on R/metadata.R:282

R/metadata.R
richfitz: no, here it means that "the name we have is final"

weshinsley (approved):

PR #116, opened 2023-11-09
mrc-4437: allow custom deserialisers in plugins

richfitz:
This feels like something that we should support, and it will be nice to get in before anyone uses a plugin in anger. I've updated the vignette as had a bunch of errors, and we now run the "full" example at last.


weshinsley (approved):
Looks good!

PR #117, opened 2023-11-20
Set wd rather than using root= in vignettes

richfitz:
As discussed in mrc-4725, try and make things a bit more obvious. I've kept the explicit temporary directory used in the initial call to `orderly_init` but added a more explicit note, then used a knitr hook to change directory for the chunks that need it. It's generally not reliable to change for the whole document; knitr will try and change back, and generated files end up in the wrong place. We already used this approach to simplify the "collaboration" vignette

### DIFF
--- a/vignettes/dependencies.Rmd
+++ b/vignettes/dependencies.Rmd
@@ -28,6 +28,16 @@ inline <- function(x) {
 
 knitr::opts_chunk$set(
   collapse = TRUE)
+
+.here <- getwd()
+knitr::knit_hooks$set(inwd = function(before, options) {
+  if (before) {
+    setwd(options$inwd)
+  } else {
+    setwd(.here)
+  }
+  invisible()
+})
 ```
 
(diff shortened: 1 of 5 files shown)


vignettes/dependencies.Rmd
```diff
@@ -28,6 +28,16 @@ inline <- function(x) {
 
 knitr::opts_chunk$set(
   collapse = TRUE)
+
+.here <- getwd()
+knitr::knit_hooks$set(inwd = function(before, options) {
```
plietar, line 33: I think it would be worth moving a lot of this block into a common file that gets sourced by all the vignettes.

plietar (approved):
LGTM, just a suggestion about moving the hook (and some existing code) to a common file.

vignettes/dependencies.Rmd
richfitz: yeah, good shout. I _think_ we can do that with `source()` - I'll give that a go and see...

PR #118, opened 2023-11-20
Simplify orderly initialisation

richfitz:
As discussed in mrc-4725, make `orderly_init` more consistent with other functions:

* use `root` as the argument, not `path`. This had been a deliberate decision but clearly not a useful one
* allow initialisation of the current directory by default (i.e., use a default arg of `"."`). This is something frowned upon by CRAN, but I can see the argument in favour

Possibly we should prevent nested repository creation (i.e., check on creation that no parent directory has .outpack) but git does not do that

### DIFF
--- a/R/example.R
+++ b/R/example.R
@@ -31,6 +31,6 @@ orderly_example <- function(name, ..., dest = NULL) {
   }
   src <- orderly2_file("example")
   fs::dir_copy(src, dest)
-  orderly2::orderly_init(..., path = dest)
+  orderly2::orderly_init(..., root = dest)
   invisible(dest)
 }
--- a/R/root.R
+++ b/R/root.R
@@ -25,8 +25,11 @@
 ##'
 ##' @title Initialise an orderly repository
 ##'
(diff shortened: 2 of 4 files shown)


plietar:
> This is something frowned upon by CRAN

I'm curious what you mean by that

plietar (approved):

richfitz:
https://cran.r-project.org/web/packages/policies.html these are the CRAN policies. Typically the first time a package is submitted (but sometimes other times) they do a manual inspection which tends to be a mixture of idiosyncratic and pedantic. The rule

> - Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.

is the one that tends to intersect with this, and in the past I and others have had demands to remove default paths that would write to useful locations, such as the working directory. It's hard to overstate how different CRAN is to crates.io/pypi etc, and how little transparency there is over decisions that are made.

PR #119, opened 2023-11-21
UX improvements for working with metadata

richfitz:
This deals with some of the metadata reading issues seen by Lydia when working with the malaria orderly2 project:

* add some interactive progress bars when building metadata
* don't read individual metadata files if we can use the data in the index (this is true most of the time)
* try and write out the index atomically (write then move, avoids corruption on network filesystems in parallel)
* in the unhappy situation where the index is corrupt, just throw it away

The progress bars are discussed here: https://cli.r-lib.org/articles/progress.html - they won't be shown unless rebuilding takes more than a couple of seconds, which is nice. They are however really annoying in tests so I've disabled them generally in the test setup and put in a pretty grim interaction test. I've tried this out Lydia's network sources, which take about 10-20s to read, and the behaviour is pretty good.

In a following PR I'll try out some way of writing out some mechanism faster than listing all files in the directory, which is still quite slow over the network

### DIFF
--- a/R/outpack_index.R
+++ b/R/outpack_index.R
@@ -72,9 +72,6 @@ index_update <- function(root_path, prev, skip_cache, progress) {
   data$unpacked <- data$location$packet[data$location$location == local]
 
   if (!identical(data, prev)) {
-    if (progress) {
-      cli::cli_progress_message("Writing index to disk")
-    }
     fs::dir_create(dirname(path_index))
     saverds_atomic(data, path_index)
   }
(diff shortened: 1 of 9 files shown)


R/outpack_index.R
```diff
@@ -55,45 +56,59 @@ outpack_index <- R6::R6Class(
   ))
 
 
-index_update <- function(root_path, prev, skip_cache) {
+index_update <- function(root_path, prev, skip_cache, progress) {
+  progress <- progress %||% getOption("orderly_index_progress", TRUE)
```
plietar, line 60: What does `%||%` do ? Does it override `progress` only when null? Is that a standard thing?

plietar (approved):

PR #120, opened 2024-01-18
Improve error message when using orderly interactively.

plietar:
It is quite easy and common for users to open up a file in RStudio and start running commands from it interactively, using Ctrl-Enter, without setting their working directory to match that file. When that happens, orderly would throw an unhelpful error, saying only `Failed to detect orderly path: /path/to/workspace` without any mention of working directories, or that the path should be a report directory (as opposed to eg. the top-level orderly workspace).

The error message is changed to look along the lines of:
```
! Working directory /path/to/workspace is not a valid orderly report.
```

This should be more helpful, by making it clear what is incorrect (the working directory), and what was expected (an orderly report).

Additionally, if running in RStudio and the current active editor is a file in what looks like an orderly report, it will augment the error message with a suggestion of what `setwd` command to run:

```
! Working directory /path/to/workspace is not a valid orderly report.
ℹ Use `setwd("/path/to/workspace/src/my-report")` to set the working directory
  to the report currently open in RStudio.
```

Finally, even if the working directory is a valid report, but it does not match the directory of the currently active editor (eg. the user has switched from one report to another), orderly commands will succeed as before, but will show a warning to the user:

```
Working directory /path/to/workspace/src/my-report does not match the report
currently open in RStudio.
ℹ Use `setwd("/path/to/workspace/src/other-report")` to switch working
  directories.
```

Frustratingly, while RStudio supports "click-to-run" snippets of code in its output, it restricts the syntax of these and in particular does not allow methods from the base package, such as `setwd`. Instead the user has to copy-paste the suggestion into the command prompt themselves. See https://github.com/rstudio/rstudio/issues/11273 and https://github.com/rstudio/rstudio/pull/11434. The restriction is pretty arbitrary and we could easily workaround it by re-exporting the `setwd` function in the orderly2 package if we really wanted to.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.10
+Version: 1.99.11
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -22,6 +22,7 @@ Imports:
     jsonlite,
     openssl,
     rlang,
+    rstudioapi,
     withr,
     yaml
 Suggests:
--- a/R/interactive.R
+++ b/R/interactive.R
@@ -21,13 +21,15 @@ rstudio_get_current_active_editor_path <- function() {
 ##
 ## I am not sure if we also want to allow working interactively from a
 ## draft directory too.
-detect_orderly_interactive_path <- function(path = getwd(),
-                                            editor_path = rstudio_get_current_active_editor_path()) {
+detect_orderly_interactive_path <- function(
+  path = getwd(),
+  editor_path = rstudio_get_current_active_editor_path())
+{
   is_valid <- is_plausible_orderly_report(path)
   suggested_wd <- NULL
   if (!is.null(editor_path)) {
     dir <- fs::path_dir(editor_path)
-    if (paths_are_different(dir, path) && is_plausible_orderly_report(dir)) {
+    if (!paths_are_identical(dir, path) && is_plausible_orderly_report(dir)) {
       suggested_wd <- dir
     }
   }
@@ -35,8 +37,9 @@ detect_orderly_interactive_path <- function(path = getwd(),
   if (!is_plausible_orderly_report(path)) {
     msg <- c("Working directory {.path {path}} is not a valid orderly report.")
     if (!is.null(suggested_wd)) {
-      cli::cli_abort(
-        c(msg, i = "Use {.code setwd({.str {suggested_wd}})} to set the working directory to the report currently open in RStudio."))
+      cli::cli_abort(c(msg, i = paste(
+        "Use {.code setwd({.str {suggested_wd}})} to set the working",
+        "directory to the report currently open in RStudio.")))
     } else {
       cli::cli_abort(msg)
     }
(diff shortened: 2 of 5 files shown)


tests/testthat/helper-orderly.R
```diff
@@ -1,6 +1,6 @@
 options(outpack.schema_validate =
           requireNamespace("jsonvalidate", quietly = TRUE) &&
-          packageVersion("jsonvalidate") >= "1.4.0",
+          utils::packageVersion("jsonvalidate") >= "1.4.0",
```
plietar, line 3: For some reason I've been getting errors on this line (even before this PR). Possibly CI never caught it because it doesn't have `jsonvalidate` installed?

plietar:
There’s a CI failure that looks unrelated, caused I think by a new version of the cli package that changed the output slightly:
https://github.com/r-lib/cli/pull/569

R/interactive.R
```diff
@@ -1,16 +1,55 @@
...
+    if (!is.null(suggested_wd)) {
+      cli::cli_abort(
+        c(msg, i = "Use {.code setwd({.str {suggested_wd}})} to set the working directory to the report currently open in RStudio."))
+    } else {
+      cli::cli_abort(msg)
+    }
+  }
+
+  if (!is.null(suggested_wd)) {
+    # For some reason, cli_warn has a different behaviour than cli_abort and
```
richfitz, line 46: I'd noticed this before; it might be worth creating an issue on r-lib/cli; Gabor is pretty receptive

R/util.R
```diff
@@ -632,3 +632,19 @@ saverds_atomic <- function(data, path, allow_fail = FALSE) {
       finally = unlink(tmp))
   }
 }
+
+
+paths_are_different <- function(x, y) {
+  tryCatch({
+    x_real <- fs::path_real(x)
```
richfitz, line 639: Any reason not to use `path_norm()` and not have it fail here?

tests/testthat/helper-orderly.R
richfitz: ci should have jsonlite installed I think, there's a Remotes line in DESCRIPTION that will get the dev version; what error were you getting?

DESCRIPTION
```diff
@@ -22,6 +22,7 @@ Imports:
     jsonlite,
     openssl,
     rlang,
+    rstudioapi,
```
richfitz, line 25: can you bump the version to 1.99.11 so we can see if the users have the version with this feature please

R/interactive.R
```diff
@@ -1,16 +1,55 @@
...
+    if (paths_are_different(dir, path) && is_plausible_orderly_report(dir)) {
+      suggested_wd <- dir
+    }
+  }
+
+  if (!is_plausible_orderly_report(path)) {
+    msg <- c("Working directory {.path {path}} is not a valid orderly report.")
+    if (!is.null(suggested_wd)) {
+      cli::cli_abort(
+        c(msg, i = "Use {.code setwd({.str {suggested_wd}})} to set the working directory to the report currently open in RStudio."))
```
richfitz, line 39: You'll need to use `paste()` to fix lint about the line length.

We've not been consistent elsewhere about `{.code ...}` etc and usually just single quote:

```
"Use 'setwd("{suggested_wd}")' to set the working directory to the report currently open in RStudio."
```

I was thinking at some point we'd go through the whole lot. I'm not convinced that showing backticks to the user might not be confusing but at the same time I don't feel strongly.

richfitz (changes requested):
Thanks, this is working well for me.

I don't mind if you fix that cli issue here or in another (or I can do it)

R/interactive.R
plietar: Sure, will do.

R/util.R
plietar: I thought maybe these could be symbolic links, and we'd want to be able to tell they are the same path. On the other hand there are probably 0 users that have symbolic links in their orderly workspaces that it is probably overkill.

tests/testthat/helper-orderly.R
plietar: Okay this turns out to just be because of my setup. I was doing a `pkgload::load_all()` inside of my RProfile to get orderly2 to load automatically, but at the point when RProfile is run, the `utils` package isn't attached yet.

I'll remove it.

R/interactive.R
plietar: https://github.com/r-lib/cli/issues/666

richfitz (approved):

PR #121, opened 2024-01-19
Fix CI issue caused by new cli and outpack versions.

plietar:
Recent versions of the cli package have added a new `vec-sep2` style option for `cli_vec`, in an effort to improve their handling of Oxford commas. This means `vec-last` is now ignored for any list of length two, and the value of `vec-sep2` is used instead.

We only used that feature in one place, with "or" as the value for `vec-last`. The new cli package broke a test by making that code print the default `vec-sep2` value, i.e. "and", when there are only two elements.

After looking at the context of that message however, I think "and" actually makes more sense than "or". It is printing a list of paths that haven't been found. We would want all of the paths in the list to be present, not just one of them. The simplest fix is therefore to remove the call to `cli_vec` and let cli use the default settings.

Finally, remove a call to `squote` on the missing paths and use a `.path` tag in the message. The end result is the same.

### DIFF
--- a/R/outpack_root.R
+++ b/R/outpack_root.R
@@ -174,9 +174,8 @@ validate_packet_has_file <- function(root, id, path, call = NULL) {
     }
   }
 
-  vmsg <- cli::cli_vec(squote(msg), list("vec-last" = " or "))
   err <- paste("Packet '{id}' does not contain the requested",
-               "{cli::qty(vmsg)} path{?s} {vmsg}")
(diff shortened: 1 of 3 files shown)


plietar:
Ah so now the coverage has changed and fails my CI because `outpack_server` can't be found because I renamed the command when I did the CLI change. I don't know if we really want to be pulling `outpack_server` from the latest commit tbh, sounds a bit brittle.

richfitz (approved):

PR #122, opened 2024-01-24
Mrc 4969 Allow <reportname>.R or orderly.R as orderly file

M-Kusumgar:
I have extracted out most instances of "orderly.R" from the code and used a util `get_orderly_file_names` to extract out the allowed file names. I have also tried to use vector methods as much as possible so that if we want to add another supported filename such as "orderly-<reportname>.R" I believe we should be able to do so by just changing that vector (and changing docs which have to have the hardcoding of file names). Let me know if that isn't the case!

Also, should I bump version number for this? And is there a `NEWS.md` file or anything like that? Can't find it.


R/metadata.R
```diff
@@ -453,8 +453,9 @@ get_active_packet <- function() {
 prevent_multiple_calls <- function(packet, name, call) {
   if (!is.null(packet$orderly2[[name]])) {
     cli::cli_abort(
-      c("Only one call to 'orderly2::orderly_{name}' is allowed",
-        i = "You have already called this function earlier in your orderly.R"),
+      c("Only one call to 'orderly2::orderly_{name}' is allowed"),
+        i = paste("You have already called this function earlier",
+                  "in your orderly file"),
```
richfitz, line 458: Can we do:
```
                  "in '{actual-name.R}'"),
```
that would be nicer if we hold that here. We should save this in the orderly-specific packet data perhaps while running?

R/run.R
```diff
@@ -266,8 +269,10 @@ orderly_run <- function(name, parameters = NULL, envir = NULL, echo = TRUE,
 
 custom_metadata <- function(dat) {
   shared <- dat$shared_resources %||% list()
+  file_names <- get_orderly_file_names(dat$src)
+  file_name <- file_names[file_exists(file_names, workdir = dat$src)]
```
r-ash, line 273: I wonder if worth getting this into the `orderly2` metadata on the packet. See line 208 - 210 in `run.R` the filename is available to you there, put it into that list and then pull it out here.

R/orderly.R
```diff
@@ -1,5 +1,5 @@
-##' List source reports - that is, directories within `src/` that
-##' contain a file `orderly.R`
+##' List source reports - that is, directories `src/<reportname>` that
+##' contain one of `<reportname>.R` or `orderly.R`
```
richfitz, line 2: Is this the point where we just disallow `orderly.R` entirely, and only have `src/<reportname>/<reportname>.R`?

R/orderly.R
```diff
@@ -21,7 +21,10 @@ orderly_list_src <- function(root = NULL, locate = TRUE) {
     return(character())
   }
   pos <- fs::dir_ls(file.path(root_path, "src"), type = "directory")
-  basename(pos)[file_exists(file.path(pos, "orderly.R"))]
+  file_paths <- lapply(pos, function(p) {
```
richfitz, line 24:
```suggestion
  file_paths <- vcapply(pos, function(p) {
```

map returning character vector I think?

R/util.R
```diff
@@ -188,15 +188,24 @@ yaml_load <- function(string) {
...
-file_exists <- function(..., workdir = NULL) {
+file_exists <- function(..., workdir = NULL, single_file = FALSE) {
   files <- c(...)
   if (!is.null(workdir)) {
     assert_scalar_character(workdir)
     owd <- setwd(workdir) # nolint
     on.exit(setwd(owd)) # nolint
   }
-  fs::file_exists(files)
+  if (single_file) {
```
richfitz, line 202: I think you have two different functions here masquerading as one with an option?

R/gitignore.R
```diff
@@ -72,7 +72,10 @@ gitignore_content_root <- function(root_path) {
 
 
 gitignore_content_src <- function(name, root_path) {
-  dat <- orderly_read_r(file.path(root_path, "src", name, "orderly.R"))
+  path <- file.path(root_path, "src", name)
+  file_names <- get_orderly_file_names(path)
+  dat <- orderly_read_r(
```
r-ash, line 77: Could just use `orderly_read` here instead?

richfitz (changes requested):
Overall, I think that the idea of name/name.R is good given the functionality of RStudio, but I am wondering if it would be better to remove the choice here.

So perhaps rework slightly to add a deprecation warning if we find `<name>/orderly.R` and at that point save into the packet runtime information the name that we found?  Then later on we can just error if the user has orderly.R

Other minor comments included, but probably quite a bit of the PR would change if you agree with this suggestion.  Curious on Rob's thoughts, and @plietar too.

r-ash (changes requested):
This looks good, but I think if you move the filename handling stuff into `orderly_read` and then return the path from that function this might simplify up some of the repeated calls to `get_orderly_file_names`

I think also worth putting this into the `orderly2` packet metadata that gets created at 208 and then using this instead of finding the filename again in `custom_metadata`.

Can you add to the vignettes too?

plietar:
> Overall, I think that the idea of name/name.R is good given the functionality of RStudio, but I am wondering if it would be better to remove the choice here.
> So perhaps rework slightly to add a deprecation warning if we find <name>/orderly.R and at that point save into the packet runtime information the name that we found? Then later on we can just error if the user has orderly.R

I think I agree with this. There's not much of a reason to support both, other than temporary backwards compatibility. I would change all the comments and documentation to just use `<name>.R` already.

R/orderly.R
```diff
@@ -74,6 +80,6 @@ orderly_new <- function(name, template = NULL, force = FALSE,
   }
 
   fs::dir_create(dest)
-  writeLines(contents, file.path(dest, "orderly.R"))
-  cli::cli_alert_success("Created 'src/{name}/orderly.R'")
+  writeLines(contents, file.path(dest, orderly_file_names[[1]]))
+  cli::cli_alert_success("Created 'src/{name}/{orderly_file_names[[1]]}'")
```
plietar, line 84: It's not obvious from looking at these two lines which alternative is actually implemented.
Assuming we decide to make the new filename the blessed option, I would just make this `src/{name}/{name}.R`.

R/orderly.R
plietar: `file_paths <- file.path(p, get_orderly_file_names(p))` works as well

M-Kusumgar:
Alright take two of this PR, I was going to use orderly context but ran into some infinite recursion in some functions and the remaining orderly context calls were throwing test errors because testthat folder is not an orderly directory (orderly context does some checks based on the working directory I think).

Either way probably best to be consistent and use the same function everywhere and do the orderly name checking there so I have used the `deprecate_old_orderly_name` function. This will throw the deprecation warning every 8 hours in the terminal if they have an `orderly.R` file.

This also includes the fix for the test that was failing due to the outpack server PR. Thanks for the help on fixing it @plietar

R/read.R
```diff
@@ -1,11 +1,12 @@
 orderly_read <- function(path, call = NULL) {
-  assert_file_exists_relative("orderly.R", name = "Orderly file",
+  orderly_name <- deprecate_old_orderly_name(path, basename(path))
+  assert_file_exists_relative(orderly_name, name = "Orderly file",
                               workdir = path, call = call)
```
plietar, line 4: Is this line necessary? Doesn't `deprecate_old_orderly_name` always return a path to an existing file?

R/gitignore.R
```diff
@@ -72,7 +72,10 @@ gitignore_content_root <- function(root_path) {
 
 
 gitignore_content_src <- function(name, root_path) {
-  dat <- orderly_read_r(file.path(root_path, "src", name, "orderly.R"))
+  src <- file.path(root_path, "src", name)
+  orderly_name <- deprecate_old_orderly_name(src, name)
+  dat <- orderly_read_r(file.path(root_path, "src", name, orderly_name),
+                        orderly_name)
```
plietar, line 78:
```suggestion
  dat <- orderly_read(file.path(root_path, "src", name))
```

Unless I'm missing something, there's no much of a reason to use `orderly_read_r` here, neither then nor nor. It's basically the same code as in that function.

R/orderly.R
```diff
@@ -49,9 +54,11 @@ orderly_new <- function(name, template = NULL, force = FALSE,
                         root = NULL, locate = TRUE) {
   root <- root_open(root, locate, require_orderly = TRUE, call = environment())
   dest <- file.path(root$path, "src", name)
+  orderly_name <- deprecate_old_orderly_name(dest, name, suppress_errors = TRUE)
+  report_name <- sprintf("%s.R", name)
```
plietar, line 58: It's a bit confusing what the two variables mean here and how they differ from one another. I think `existing_report_file` and `new_report_file` would better explain the difference.

R/read.R
```diff
@@ -1,11 +1,12 @@
...
-  orderly_read_r(file.path(path, "orderly.R"))
+  orderly_read_r(file.path(path, orderly_name), orderly_name)
 }
 
 
-orderly_read_r <- function(path) {
+orderly_read_r <- function(path, orderly_name) {
```
plietar, line 9:
```suggestion
orderly_read_r <- function(path) {
```

Argument isn't being used anywhere.

plietar (commented):
Looks good. Just some minor nitpicks:
- `orderly_name`, as are called a few variables, is quite a confusing, erm, name. It could mean anything really. I think something like `script_name` or `entrypoint_filename` is much clearer.
- `deprecate_old_orderly_name` is also confusing. While it does indeed print a depreciation note, that's kind of a side effect. The main thing it does and why we call it is to find the name of the file. `locate_report_file` or `find_entrypoint_file` or something along those lines would be better IMO.
- I think `orderly_read` should include the file's name in the returned dictionary. The function reads a bunch of metadata about the report, and could include that name. Most users of `deprecate_old_orderly_name` would have just called `orderly_read` a few lines prior. There are a few cases where that's not the case, and these places just call `deprecate_old_orderly_name` directly and that is fine IMO.

R/read.R
M-Kusumgar: this is being used for ret$resources line in the function towards the end of it

R/read.R
M-Kusumgar: no the assert file exists check is not necessary anymore!

R/gitignore.R
M-Kusumgar: true yh we could just have `orderly_read` do the deprecation check, done

M-Kusumgar:
> * `orderly_name`, as are called a few variables, is quite a confusing, erm, name. It could mean anything really. I think something like `script_name` or `entrypoint_filename` is much clearer.

Agreed, I like `entrypoint_filename` so have gone with that

> * `deprecate_old_orderly_name` is also confusing. While it does indeed print a depreciation note, that's kind of a side effect. The main thing it does and why we call it is to find the name of the file. `locate_report_file` or `find_entrypoint_file` or something along those lines would be better IMO.

For consistency in the naming I have gone with `find_entrypoint_filename`

> * I think `orderly_read` should include the file's name in the returned dictionary. The function reads a bunch of metadata about the report, and could include that name. Most users of `deprecate_old_orderly_name` would have just called `orderly_read` a few lines prior. There are a few cases where that's not the case, and these places just call `deprecate_old_orderly_name` directly and that is fine IMO.

I found one example in `orderly_run` have returned `entrypoint_filename` key in list from `orderly_read_r` and used this in there, not sure where else I would use it, was there another place you had in mind?

R/gitignore.R
```diff
@@ -72,7 +72,8 @@ gitignore_content_root <- function(root_path) {
 
 
 gitignore_content_src <- function(name, root_path) {
-  dat <- orderly_read_r(file.path(root_path, "src", name, "orderly.R"))
+  src <- file.path(root_path, "src", name)
```
plietar, line 75:
```suggestion
```

R/read.R
plietar: Yeah, missed that sorry

plietar (approved):

R/metadata.R
```diff
@@ -454,7 +454,8 @@ prevent_multiple_calls <- function(packet, name, call) {
   if (!is.null(packet$orderly2[[name]])) {
     cli::cli_abort(
       c("Only one call to 'orderly2::orderly_{name}' is allowed",
-        i = "You have already called this function earlier in your orderly.R"),
+        i = paste("You have already called this function earlier",
+                  "in your <reportname>.R")),
```
richfitz, line 458: Pass the filename down here to `prevent_multiple_calls` and use it in the error message, it'll be clearer.

R/orderly.R
```diff
@@ -1,5 +1,5 @@
 ##' List source reports - that is, directories within `src/` that
-##' contain a file `orderly.R`
+##' contain a file `<reportname>.R`
```
richfitz, line 2: I think this could be reworded or explained better

```suggestion
##' look suitable for running with orderly; these will be directories
##' that contain a `.R` file with the same name as the directory
##' (e.g., `src/data/data.R` corresponds to `data`).
```

More of a mouthful, but perhaps clearer?

R/util_assert.R
```diff
@@ -80,6 +80,37 @@ assert_file_exists <- function(files, name = "File", call = NULL, arg = NULL) {
...
+find_entrypoint_filename <- function(src, reportname,
+                                       suppress_errors = FALSE) {
+  names <- c(sprintf("%s.R", reportname), "orderly.R")
+  files_exist <- file.exists(file.path(src, names))
+  if (sum(files_exist) > 1 && !suppress_errors) {
+    cli::cli_abort(
+      paste("Please only create {names[[1]]} file, orderly.R",
+            "has been deprecated")
+    )
+  } else if (sum(files_exist) == 0 && !suppress_errors) {
```
richfitz, line 92: Our normal style here is not to have an else clause after an if that throws.  So

```
if (whatever) {
  throw
}
otherwise
```

not

```
if (whatever) {
  throw
} else {
  otherwise
}
```

R/orderly.R
```diff
@@ -49,9 +54,13 @@ orderly_new <- function(name, template = NULL, force = FALSE,
                         root = NULL, locate = TRUE) {
   root <- root_open(root, locate, require_orderly = TRUE, call = environment())
   dest <- file.path(root$path, "src", name)
+  existing_entrypoint_filename <- find_entrypoint_filename(
```
richfitz, line 57: I don't think this is quite correct, because you'll mishandle the case where we have `src/foo/foo.R` and `src/foo/orderly.R` so we return `NULL` and think we can continue and then overwrite the existing file.

R/read.R
```diff
@@ -1,11 +1,10 @@
 orderly_read <- function(path, call = NULL) {
-  assert_file_exists_relative("orderly.R", name = "Orderly file",
-                              workdir = path, call = call)
-  orderly_read_r(file.path(path, "orderly.R"))
+  entrypoint_filename <- find_entrypoint_filename(path, basename(path))
```
richfitz, line 2: Almost every call here has this pattern. Consider refactoring so that you can do

```
find_entrypoint_filename(base)
```

(i.e., move the `basename` within `find_entrypoint_filename`) which will make this a little less noisy I think.

R/run.R
```diff
@@ -546,6 +551,10 @@ validate_orderly_directory <- function(name, root_path, call) {
     cli::cli_abort(err, call = call)
   }
 
+  find_entrypoint_filename(
```
richfitz, line 554: this is just being called for the warning/error? A comment perhaps? Will this be removed after we remove the deprecation warning?

R/util_assert.R
```diff
@@ -80,6 +80,37 @@ assert_file_exists <- function(files, name = "File", call = NULL, arg = NULL) {
...
+      .frequency = "regularly",
+      .frequency_id = "deprecate_orderly_file_name"
+    )
+  }
+
+  if (suppress_errors && sum(files_exist) != 1) {
+    NULL
+  } else {
+    names[files_exist]
+  }
```
richfitz, line 110:
```suggestion
  if (sum(files_exist) == 1) names[files_exist] else NA_character_
```

* no need to test `suppress_errors`
* see above for suggestion to replace return type with missing string rather than `NULL`
* can do it on one line, which I find sometimes clearer

Also you sum files_exist several times, here could pull that into `n_found <- sum(files_exist)` if you want

_pkgdown.yml
```diff
@@ -28,7 +28,7 @@ reference:
       - orderly_list_src
   - title: From within a running report
     desc: >-
-      These are the functions that get called from your `orderly.R` file
+      These are the functions that get called from your `<reportname>.R` file
```
richfitz, line 31: I think we need a nicer name for this.  Perhaps calling it your "orderly file" or "entrypoint file" everywhere would make it nicer to read?

tests/testthat/examples/computed-resource/computed-resource.R
richfitz, line 1: In another PR we're going to need DYM support here for people getting things like underscores and hyphens confused (they do this quite a lot already)

R/orderly.R
richfitz: should be able to verify this with your two-orderly-files test?

vignettes/dependencies.Rmd
```diff
@@ -28,7 +28,7 @@ Here, we show how to practically use dependencies in a few common scenarios of i
 
 ## Basic use
 
-The primary mechanism for using dependencies is to call `orderly2::orderly_dependency()` from within an `orderly.R` script; this finds a suitable completed packet and copies files that are found from within that packet into your current report.
+The primary mechanism for using dependencies is to call `orderly2::orderly_dependency()` from within a `<reportname>.R` script; this finds a suitable completed packet and copies files that are found from within that packet into your current report.
```
richfitz, line 31: yeah, we need a name for this!

richfitz (changes requested):
Thanks - small suggestions here

R/read.R
M-Kusumgar: done

R/run.R
M-Kusumgar: yep added a comment, this is just for the deprecation warning

R/metadata.R
M-Kusumgar: done, just did it using the packet info we have, thought that was simpler

R/util_assert.R
M-Kusumgar: yep done!

R/util_assert.R
M-Kusumgar: yh looks much cleaner, done

R/orderly.R
M-Kusumgar: yh definitely clearer, ive altered it slightly, have said it will list dirs that contain an entrypoint file - a `.R` file with the same name...

_pkgdown.yml
M-Kusumgar: yh i think ill do entrypoint file? will be consistent with our variable naming!

M-Kusumgar: although entrypoint file is a bit difficult to understand out of context, maybe for docs we go for orderly file? might be clearer for the users

R/util_assert.R
```diff
@@ -80,6 +80,37 @@ assert_file_exists <- function(files, name = "File", call = NULL, arg = NULL) {
...
+    cli::cli_abort(
+      paste("Please only create {names[[1]]} file, orderly.R",
+            "has been deprecated")
+    )
+  } else if (sum(files_exist) == 0 && !suppress_errors) {
+    cli::cli_abort(
+      "Please create {names[[1]]} file"
+    )
+  } else if (files_exist[[2]]) {
+    rlang::inform(
```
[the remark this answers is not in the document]
M-Kusumgar: done

tests/testthat/examples/computed-resource/computed-resource.R
M-Kusumgar: created a ticket: https://mrc-ide.myjetbrains.com/youtrack/agiles/103-58/current?issue=mrc-5100

richfitz (changes requested):
Looking good.

> Also, should I bump version number for this? And is there a NEWS.md file or anything like that? Can't find it.

Just bump the patch version number for now

r-ash (approved):

PR #123, opened 2024-01-31
add installation to introduction

M-Kusumgar:
Guess this one is up for debate. I realised we have installation guide on the README which is the homepage for the docs but I still stupidly clicked on introduction and was like how do I install orderly2.


richfitz (approved):
this is great thanks - we have used that pattern in other docs.

PR #125, opened 2024-02-07
Use existing files on disk instead of pushing afresh

r-ash:
Noticed this when doing the Python impl, in particular here https://github.com/mrc-ide/outpack-py/pull/33#discussion_r1481253616

We have logic which checks if a file is already known on disk somewhere and then adds this into a temporary filestore. We are not then removing that from the list of files to fetch from the location. I've added a test which checks this case and updated this

Note this is failing due to an outpack_server change, Mantra has included the fix for this on his PR #122 so we should merge this after that

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.12
+Version: 1.99.13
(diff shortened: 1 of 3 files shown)


richfitz (changes requested):
Sorry

richfitz (approved):

PR #129, opened 2024-02-29
Add function to return details of report script from parsed exprs

r-ash:
In the runner we want to get the parameters for a report given a name and a hash to run on. So we use git show to see the contents of the file at this point. So we want a way to pass this already read contents into this function to extract the details (so we can look at the parameters)

Struggling to think of a great name for the function here so very open for any suggestions, also wondering if I should do a `@noRd` on this so that we're not advertising this as a function for other people to use?


NAMESPACE
```diff
@@ -4,6 +4,7 @@ S3method(format,orderly_query)
 S3method(print,orderly_cleanup_status)
 S3method(print,orderly_query_explain)
 export(orderly_artefact)
+export(orderly_build_script_details)
```
richfitz, line 7: What about calling it `orderly_parse()` and have it take arguments of _either_ `path` or `expr`, then just documenting it as experts use? I can see this being generally useful really

richfitz (changes requested):

NAMESPACE
r-ash: Yep I like that

R/read.R
```diff
@@ -1,11 +1,30 @@
...
+#' of any calls to the orderly_ in-script functions into intermediate
+#' representation for downstream use. Also validates that any calls to
+#' orderly_ in-script functions are well-formed.
+#'
+#' @param entrypoint_script Path to script or parsed AST from orderly script
+#' @param entrypoint_filename Name of entrypoint file to include in metadata
+#'
+#' @return Parsed orderly entrypoint script
+#' @export
+orderly_parse <- function(entrypoint_script, entrypoint_filename) {
```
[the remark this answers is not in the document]
r-ash: Initially put this in but felt kind of gross when chatting through it with Mantra, I've split this into 2 functions now. How does that look?

R/read.R
```diff
@@ -1,11 +1,36 @@
...
+orderly_parse_file <- function(path) {
+  assert_file_exists(path)
   exprs <- parse(file = path)
+  orderly_parse_expr(exprs, basename(path))
+}
+
+
+#' @param exprs Parsed AST from `orderly_*` script
+#' @param filename Name of `orderly_*` file to include in metadata, required if
+#'   calling with `exprs`. If called with a `path` this can be NULL.
```
M-Kusumgar, line 28: i think might've been from before?
```suggestion
#' @param exprs Parsed AST from `orderly_*` script
#' @param filename Name of `orderly_*` file to include in metadata
```

M-Kusumgar (approved):
Nice one! i think tiny doc change but otherwise all good

richfitz (approved):

PR #131, opened 2024-03-04
Add support for authenticating against packit

richfitz:
Implement support for using packit as a location for orderly2.

There were two broad ways I thought we might do this; one is that we say that packit locations are merely http ones and we add support for optional auth and the users have to give the url with `/packit/api/outpack`, but this feels like the users will have to do remember to add a url that looks quite different to the one they feel it should be.

So I've gone with a new location type "packit".  This turns out to be useful because the login url is actually above the location where we refer to outpack

```
/packit/api/auth/login/api         the login url
/packit/api/outpack/metadata/list  example api url
```

This approach leaves the configuration fairly tidy I think, and requires no schema updates.

Otherwise not that much to see in the implementation:

* I've added in some basic slash sanitisation for both http and packit remotes (it should not matter if there is or is not a trailing slash)
* the token can come from an envvar; we might support using the keychain later but that's more work
* login is very slow so I've made a global cache of logins against a hash of auth data
* packit produces errors in slightly different format to outpack; we should harmonise these at some point
* in orderlyweb (for orderly1) we refreshed the token on timeout; that can be handled fairly easily (see comment in `http_client_handle_error`)
* the testing is all gross and mocky

I can run this up locally with:

```r
tmp <- tempfile()
orderly_init(tmp)
orderly_location_add(
  "malaria",
  "packit",
  list(url = "https://malaria-orderly.dide.ic.ac.uk/",
       token = "$VAULT_AUTH_GITHUB_TOKEN"),
  root = tmp)
orderly_location_list(root = tmp)
orderly_location_pull_metadata(root = tmp)
```

Subsequent calls to `orderly_location_pull_metadata` won't log back in, and will reuse the jwt

### DIFF
--- a/R/outpack_http_client.R
+++ b/R/outpack_http_client.R
@@ -1,33 +1,50 @@
-## NOTE: none of the auth bits here are done yet - we have a system in
-## the orderlyweb client that lets us do this in a fairly pluggable
-## way, supporting username/password and token based auth (with the
-## former getting a token via username/password)
 outpack_http_client <- R6::R6Class(
   "outpack_http_client",
 
+  private = list(
+    auth = NULL
+  ),
+
   public = list(
     url = NULL,
+    token = NULL,
 
-    initialize = function(url) {
+    initialize = function(url, auth) {
       self$url <- url
+      if (is.null(auth)) {
+        private$auth <- list(enabled = FALSE)
+      } else {
+        private$auth <- list(enabled = TRUE, url = auth$url, data = auth$data)
+      }
+    },
+
+    authorise = function() {
+      needs_auth <- private$auth$enabled && is.null(private$auth$header)
+      if (needs_auth) {
+        private$auth$header <- http_client_login(self$url, private$auth)
+      }
     },
 
     get = function(path, ...) {
-      http_client_request(httr::GET, self$url, path, ...)
+      self$authorise()
+      http_client_request(httr::GET, paste0(self$url, path), ...,
+                          private$auth$header)
     },
 
     post = function(path, body, ...) {
-      http_client_request(httr::POST, self$url, path, body = body, ...)
+      self$authorise()
+      http_client_request(httr::POST, paste0(self$url, path), body = body, ...,
+                          private$auth$header)
     }
   ))
 
 
-http_client_request <- function(verb, server, path, ..., parse_json = TRUE,
+http_client_request <- function(verb, url, ..., parse_json = TRUE,
                                 download = NULL) {
   if (is.null(download)) {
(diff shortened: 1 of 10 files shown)


R/outpack_http_client.R
```diff
@@ -71,3 +87,28 @@ http_client_download_options <- function(dest) {
...
+http_client_login <- function(name, auth) {
+  key <- rlang::hash(auth)
+  if (is.null(auth_cache[[key]])) {
+    cli::cli_alert_info("Logging in to {name}")
+    res <- http_client_request(httr::POST, auth$url,
+                               body = auth$data, encode = "json")
+    cli::cli_alert_success("Logged in successfully")
+    auth_cache[[key]] <- httr::add_headers(
+      "Authorization" = paste("Bearer", res$token))
+  }
```
r-ash, line 112: Do we know how long the jwt tokens are valid for? If we get an error on request about an expired token we could clear the cache and retry the request with the token. Probably not a big issue, we might be unlikely to encounter it if the tokens are long living enough and easy to fix through manually clearing your cache or restarting the session.

r-ash (approved):
This looks good, looks like it should be fine to port to Python too. 

Bit of an aside but I wonder if we should add a function `outpack_location_info` which tells you about the locations you have configured. At the moment very easy to add a location with some poorly chosen name and forget where that is actually pushing and pulling from.

richfitz:
> Bit of an aside but I wonder if we should add a function outpack_location_info which tells you about the locations you have configured. At the moment very easy to add a location with some poorly chosen name and forget where that is actually pushing and pulling from.

Agreed; I had same thought when implementing this and added: https://mrc-ide.myjetbrains.com/youtrack/issue/mrc-5130/orderlylocationlist-verbose

R/outpack_http_client.R
richfitz: yeah, this is what I was thinking about with this comment:

> in orderlyweb (for orderly1) we refreshed the token on timeout; that can be handled fairly easily (see comment in http_client_handle_error)

This is how we handle it in orderlyweb: https://github.com/vimc/orderlyweb/blob/master/R/api_client.R#L111-L118

@EmmaLRussell ; do you know how Packit handles this? Or do you know how long the jwts are valid for?

EmmaLRussell: Jwt expiry time is configurable on deploy for packit: https://github.com/mrc-ide/packit/blob/main/api/app/src/main/resources/config.properties#L10

richfitz: amazing thanks - details now in mrc-5160 and will sort this out in another PR

PR #132, opened 2024-03-18
Add functions to export and import packets to/from a zip file

plietar:
While the ability to push and pull from remote locations that orderly
already supports works well for back-and-forth collaboration, it is not
as well suited to produce a one-time artefact that may be released
and shared publicly, without depending on a server or shared location.
One use case for this is for publishing a reproducible set of analyses
to accompany a paper.

This commit adds a pair of functions, orderly_export_zip and
orderly_import_zip. These allow a set of packets (and their transitive
dependencies) to be exported as a standalone zip file, containing both
the metadata and files. The zip file can then be imported into a
different repository.

The zip file is formatted as a metadata directory, with a file per
packet, and a content-addressed file store.

### DIFF
--- a/R/zip.R
+++ b/R/zip.R
@@ -67,16 +67,15 @@ orderly_zip_import <- function(path, root = NULL, locate = TRUE) {
   root <- root_open(root, locate = locate, require_orderly = FALSE,
                     call = environment())
 
-  src <- withr::local_tempfile()
-  zip::unzip(path, exdir = src)
-
-  if (!file.exists(file.path(src, "outpack.json"))) {
+  if (!("outpack.json" %in% zip::zip_list(path)$filename)) {
     cli::cli_abort(
       c("Zip file does not contain an 'outpack.json' file at its root",
         i = paste("Are you sure this file was produced by",
-                  "orderly2::orderly_zip_export?")),
-      call = environment())
+                  "orderly2::orderly_zip_export?")))
   }
+
+  src <- withr::local_tempfile()
+  zip::unzip(path, exdir = src)
 
   contents <- jsonlite::read_json(file.path(src, "outpack.json"),
                                   simplifyVector = TRUE)


plietar:
There's an open question left in the implementation, which I've marked as a TODO: when importing the zip file, we need to hash the metadata files to add the packet to the `local` location. At present, I've hashed the metadata file using the algorithm configured on the root we are importing into. However this is wrong if the root that originally created the packet used a different algorithm. We might end up with two different hashes, using two different algorithms, for the same packet id. This will cause problems later if the two roots end up sharing a location.

To solve this we would need to record the hash (or hash algorithm) somewhere in the zip file. I'm not sure what the best place would be. Possible options:
- Include an index file at the root of the zip file, with the list of packets and their hash. This is more or less the equivalent of the `/metadata/list` endpoint of the http API.
- Wrap each metadata files in a JSON object, along the lines of `{ "hash": "sha256:xxx", "contents": "{ ... }" }`, where contents is a string containing the packet metadata. We need it to be a string to preserve its encoding and hash.
- Name the metadata files based on their hash. Basically this would make the metadata repository content-addressed as well. We wouldn't loose track of the packet ID since it is contained in the metadata.

plietar:
I ended up going with the first option, of including an `outpack.json` file at the root of the zip file which contains the list of packets and their hash. This also lets us detect zip files that are obviously not supposed to be imported.

R/export.R
```diff
@@ -0,0 +1,149 @@
+##' Export packets as a zip file.
+##'
+##' The packets can be imported into a different repository using the
+##' [orderly2::orderly_import_zip] function.
+##'
+##' This is useful as one-time way to publish your results, for example as an
+##' artefact accompanying a paper. For back-and-forth collaboration, a shared
+##' location should be priviledged.
```
richfitz, line 8:
```suggestion
##' location should be preferred, as this offers more flexibility.
```

R/export.R
```diff
@@ -0,0 +1,149 @@
...
+##'
+##' @param path the path where the zip file will be created
+##'
+##' @param packets One or more packets to export
+##'
+##' @inheritParams orderly_metadata
+##'
+##' @return Nothing
```
richfitz, line 16: I suggest invisibly returning `path`

R/export.R
```diff
@@ -0,0 +1,149 @@
...
+    match(index$location$packet, packets), c("packet", "hash")]
+  contents <- list(packets = packet_list)
+
+  writeLines(to_json(contents, "orderly/export.json"),
+             file.path(dest, "outpack.json"))
+
+  zip::zip(fs::path_abs(path), root = dest,
+           files = c("outpack.json", "metadata", "files"))
+
+  invisible()
```
richfitz, line 55:
```suggestion
  invisible(path)
```

R/export.R
```diff
@@ -0,0 +1,149 @@
...
+##' @inheritParams orderly_metadata
+##'
+##' @return Nothing
+##' @export
+orderly_export_zip <- function(path, packets, root = NULL, locate = TRUE) {
+  root <- root_open(root, locate = locate, require_orderly = FALSE,
+                    call = environment())
+
+  index <- root$index$data()
+  packets <- find_all_dependencies(packets, index$metadata)
```
richfitz, line 23: I wonder if we should offer a non-recursive option here at some (future) point? So that you can dump out a particular packet but _not_ all the bits that lead to it? This would be much smaller often, and more relevant to most users.

The issue here is this requirement:

> She has a sequence of Orderly reports. The first report pulls in from a private data source (which cannot be shared), does a bit of analysis and produces an artefact. The subsequent reports build from the first one's output and apply further analyses.

So by not including all dependencies then we can be sure that no private data has been exposed. 

At present we need to include all metadata I believe, but importing a packet into a location without its dependent files should be fine so long as the user has `require_complete_tree` set to `FALSE`

R/export.R
```diff
@@ -0,0 +1,149 @@
...
+##' @return Invisibly, the IDs of the imported packets
+##' @export
+orderly_import_zip <- function(path, root = NULL, locate = TRUE) {
+  root <- root_open(root, locate = locate, require_orderly = FALSE,
+                    call = environment())
+
+  src <- withr::local_tempfile()
+  zip::unzip(path, exdir = src)
+
+  if (!file.exists(file.path(src, "outpack.json"))) {
```
richfitz, line 73: You can fail faster with `zip::zip_list` here perhaps?

_pkgdown.yml
```diff
@@ -66,6 +66,10 @@ reference:
       - orderly_location_list
       - orderly_location_remove
       - orderly_location_rename
+  - title: Exporting packets
+    contents:
+      - orderly_export_zip
+      - orderly_import_zip
```
richfitz, line 72: Torn here on the names; I wonder about `orderly_zip_import` / `orderly_zip_export` as most of the names work like this (to help with autocomplete). OTOH, if we had different import/export options in future, your way is better.

richfitz (changes requested):
Thanks Paul - this is great, and I think some of the issues that I raised here will sort themselves out once it collides with reality.

The biggest bit to think about is whether we want the whole tree or not, and if we want to think about this now or later. We've actually done more extreme versions of this in the COVID work where we needed to create a new packet that "starts again", cutting itself off from its parents with all artefacts appearing fully formed (probably this will be easier to explain with waving hands).

From the ticket, I do think that the static site version might be worth thinking about - it's not that bad to set up a site, though we may find that users hit issues with what git (and before that github) likes to store.  Use of the github api to use releases might help there though, as that's a great place to throw bigger files.

R/export.R
plietar: I hadn't considered the fact that orderly might be storing the inputs to the first task. I think she is using `orderly_shared_resource` to ingest the data, which is probably going to leak into the zip file.

On the other hand, not including the first packet means you can't run the reproduce the second one, at least not without some shenanigans to extract the inputs from the second packet.

I think we might need a way of including the first packet, but only its artefacts, not its inputs.

I'll check with Sangeeta exactly how she'd doing it to make sure nothing is leaking accidentally.

_pkgdown.yml
plietar: Yeah now that you mention it I think I like those names better. I renamed them, and also renamed the file to `zip.R`.

I think I would have preferred a more generic name instead of `zip`, and then in the future accept a `format = "zip"` argument that could take other values. Sadly the most obvious candidate, `archive`, is already used to mean something completely different. `tarball` kind of works, but it implies the `tar` format which is not much better than `zip` is.

R/export.R
plietar: Nevermind, I was wrong about where the data comes from. It actually comes from an Microsoft Access database, and the first report opens up a connection and retrieves the data it needs. There shouldn't be any particular concern about data leakage, since anything that is fetched from the database can be made public.

We might need to do something different in the future, but this seems fine for now.

R/export.R
plietar: Done.

PR #134, opened 2024-03-25
Use lockfile when installing outpack_server in CI.

plietar:
Somewhere, something in the dependency tree of outpack_server something changes and caused the build to fail, which broke the orderly2 CI.

This is the kind of thing lock files are supposed to prevent, by pinning to the same package versions used in outpack_server's own CI, which should make the build reproducible. However, quite surprisingly, `cargo install` ignores the lock file by default and pulls the latest compatible version of dependencies. Passing the `--locked` argument should prevent this from happening.

I haven't investigated the underlying build failure much yet, but seems like an issue with a missing cargo feature flag.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -52,7 +52,7 @@ jobs:
       - name: setup server
         if: runner.os != 'macOS'
         run: |
-          cargo install --git https://github.com/mrc-ide/outpack_server --branch main
+          cargo install --git https://github.com/mrc-ide/outpack_server --branch main --locked
 
       - uses: r-lib/actions/setup-r-dependencies@v2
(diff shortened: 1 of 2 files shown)


plietar:
Previous CI failure: https://github.com/mrc-ide/orderly2/actions/runs/8421643492/job/23059035785?pr=132

In the logs it says `Downloaded tokio v1.36.0`, whereas the outpack_server lockfile has it set to 1.35.1: https://github.com/mrc-ide/outpack_server/blob/dbf23fb4f6a84cc326efd48c1011816a181e5dfd/Cargo.lock#L2174. I don't think tokio is the actual dependency at fault here (1.36.0 was released over a months ago, and this issue on started today), just the first one I noticed that had a version mismatch,

PR #136, opened 2024-04-10
Allow orderly_shared_resource to be used with unnamed arguments.

plietar:
The previous behaviour was inconsistent with orderly_resource and often redundant, as users commonly want to copy the resource using the same name as the original file.

It is now possible to use `orderly_shared_resource("foo.txt")`, in which case the source and destination name are assumed to be the same.

The old syntax is still available in case the user wants to change the name while copying, and it is even possible to mix-and-match, renaming some files but not others.

### DIFF
--- a/R/metadata.R
+++ b/R/metadata.R
@@ -325,9 +325,9 @@ static_orderly_dependency <- function(args) {
 ##'
 ##' @title Copy shared resources into a packet directory
 ##'
-##' @param ... Named arguments corresponding to shared resources to
-##'   copy. The name will be the destination filename, while the value
-##'   is the filename within the shared resource directory.
+##' @param ... The shared resources to copy. If arguments are named, the name
+##'   will be the destination file while the value is the filename within the
+##'   shared resource directory.
 ##'
 ##' @return Invisibly, a data.frame with columns `here` (the fileames
 ##'   as as copied into the running packet) and `there` (the filenames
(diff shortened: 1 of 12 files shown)


R/metadata.R
```diff
@@ -351,21 +351,23 @@ orderly_shared_resource <- function(...) {
...
 }
 
-
 validate_shared_resource <- function(args, call) {
   if (length(args) == 0) {
     cli::cli_abort("'orderly_shared_resource' requires at least one argument",
                    call = call)
   }
-  assert_named(args, unique = TRUE, call = environment())
+
```
richfitz, line 359: Can you see if you can harmonise this through `validate_file_from_to` (in outpack_misc.R) which contains most of this logic?  I think it should just drop in directly?

tests/testthat/test-run.R
```diff
@@ -261,21 +261,54 @@ test_that("can run manually with shared resources", {
...
-    c(a = "A", b = "B"))
-})
+    orderly_shared_resource(a = 1, b = TRUE, c = "str"),
+    "All elements of 'arguments to 'orderly_shared_resource'' must be strings")
+  expect_error(
+    orderly_shared_resource(1, TRUE, "str"),
+    "All elements of 'arguments to 'orderly_shared_resource'' must be strings")
+  expect_error(
+    orderly_shared_resource(a = 1, TRUE, "str"),
+    "All elements of 'arguments to 'orderly_shared_resource'' must be strings")
```
plietar, line 297: These messages are pretty terrible, but I couldn't find a way to deparse the `...` reliably so I've just used a placeholder instead.

I'd almost be tempted to change `orderly_shared_resource` to accept a single files argument instead, making it consistent with all the other functions, including `orderly_dependency`, `orderly_copy_files` and `orderly_resource`, but of course that's a breaking change.

PR #140, opened 2024-05-30
Better handling of failed push due to corrupt packets

richfitz:
This does not actually **cope** with this situation any better, but it prints the hashes and does not error while reporting the error, which was the previous behaviour.


richfitz:
This is failing on windows due to building the rust binary there :(

plietar (approved):

Issue #144, opened 2024-06-18
orderly_location_pull_packet() reports "✔ All files available locally, no need to fetch any" when it doesn't find anything


jeffeaton:
I am finding that `"orderly_location_pull_packet()"` returns message `✔ All files available locally, no need to fetch any`, even when it doesn't find anything.

```r
> orderly_location_pull_packet(name = "hello")
ℹ Looking for suitable files already on disk
✔ All files available locally, no need to fetch any
> orderly_metadata_extract()  ## Returns nothing
[1] id         name       parameters
<0 rows> (or 0-length row.names)
```

This happens often if I forget to pull metadata before trying to pull a packet. I am unsure if this is intentional behaviour. I would find it more useful to return a message `"I didn't find the packet you requested"`.

Here is little example I created:

```r
library(orderly2)

root1 <- tempfile()
root2 <- tempfile()

## * Create orderly instance at root1
## * Create a report
## * Run the report to create package in archive
orderly_init(root1)

setwd(root1)
orderly_new("hello")

writeLines('
library(orderly2)
orderly_artefact("hello", "hello.txt")
writeLines("hello world!", "hello.txt")
',
"src/hello/hello.R"
)

id <- orderly_run("hello")


## * Create orderly instance at root2
## * Add root1 as a location in root2

orderly_init(root2)
setwd(root2)

orderly_location_add("root1", "path", list(path = root1))

## This is surprising: if I try to pull packet "hello" from
## root1 without first pulling the metadata, it says
## "✔ All files available locally, no need to fetch any"
## even thought it should have found nothing.

orderly_location_pull_packet(name = "hello")

orderly_metadata_extract()  ## Returns nothing

## Proper thing to do is (1) pull metadata, then (2) pull packet
orderly_location_pull_metadata(location = "root1")
orderly_location_pull_packet(name = "hello")

orderly_metadata_extract()  ## Lists the packet
```

Session info: 
```
> sessionInfo()
R version 4.3.2 (2023-10-31)
Platform: aarch64-apple-darwin20 (64-bit)
Running under: macOS Ventura 13.6

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRblas.0.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.11.0

locale:
[...]

PR #145, opened 2024-06-24
mrc-5483: Fix issue when pulling multiple packets with multiple files

r-ash:
Issue spotted by Jeff when pulling packets

```
> orderly2::orderly_location_pull_packet(name = "split_model_results")
Error in data.frame(..., stringsAsFactors = FALSE, check.names = FALSE) : 
  arguments imply differing number of rows: 537, 31

> traceback()
6: stop(gettextf("arguments imply differing number of rows: %s", 
       paste(unique(nrows), collapse = ", ")), domain = NA)
5: data.frame(..., stringsAsFactors = FALSE, check.names = FALSE)
4: data_frame(hash = unlist(lapply(meta, function(x) x$files$hash), 
       FALSE, FALSE), size = unlist(lapply(meta, function(x) x$files$size), 
       FALSE, FALSE), location = location_use)
3: location_build_pull_plan_files(packets$fetch, location, root, 
       call)
2: location_build_pull_plan(ids, options$locations, recursive, root, 
       call = environment())
1: orderly2::orderly_location_pull_packet(name = "split_model_results")
```

This is happening because if we pull with a query which finds multiple packets which are at more than 1 location then we're not matching up the location to the file information correctly. In particular, if the packets have a total number of files which is not a multiple of the number of locations then this errors :see_no_evil: Expect we've not seen this in our other tests because we're always creating 1 file in the test data. This PR adds a test to recreate the issue and fixes it.

I tidied up the docs here a little bit too. But Jeff was definitely a bit confused about this interface, he was expecting passing `name = "..."` to only pull the latest with that name. It wasn't immediately obvious to him that this would pull all of them, and I can see why. Will leave this for @M-Kusumgar to discuss

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -273,7 +273,7 @@ orderly_location_pull_metadata <- function(location = NULL, root = NULL,
 ##' efficient, as we keep track of files that are copied over even in
 ##' the case of an interrupted pull.
 ##'
-##' @title Pull a single packet from a location
+##' @title Pull one or more packets from a location
 ##'
 ##' @param ... Arguments passed through to
 ##'   [orderly2::orderly_search]. In the special case where the first
--- a/R/location.R
+++ b/R/location.R
@@ -673,10 +673,9 @@ location_build_pull_plan_files <- function(packet_id, location, root, call) {
         intersect(location, loc$location[loc$packet == id])[[1]]
       }, USE.NAMES = FALSE)
     }
-    files <- data_frame(
-      hash = unlist(lapply(meta, function(x) x$files$hash), FALSE, FALSE),
-      size = unlist(lapply(meta, function(x) x$files$size), FALSE, FALSE),
-      location = location_use)
+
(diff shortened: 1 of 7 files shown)


richfitz:
Mantra and I discussed adding a `latest = ` helper here too, in the same way as `name` is a helper

richfitz:
This issue also turns up in #144 and #146 implicitly in that we might want a "pull exactly one" vs "pull all that match" interface

richfitz (approved):

R/location.R
```diff
@@ -273,7 +273,7 @@ orderly_location_pull_metadata <- function(location = NULL, root = NULL,
 ##' efficient, as we keep track of files that are copied over even in
 ##' the case of an interrupted pull.
 ##'
-##' @title Pull a single packet from a location
+##' @title Pull one or more packets from a location
```
richfitz, line 276: clearly our idea and implementation have drifted apart at some point

PR #147, opened 2024-06-25
Add a function to compare the contents of two packets.

plietar:
The new `orderly_compare_packets` function takes in two packet IDs and produces a diff of the two packets' contents. This includes a comparison of both the packets' metadata and of their files. The actual diff of the files is computed using the diffobj package.

A parameter of the function can be used to control which components of the packets are considered in the comparison, and take be one of "everything", "metadata", "files" and "artefacts".

This feature was requested as part of the epireview project, where one report's implementation got very messy and needs refactoring, but in the absence of any way of comparing two runs of the same report it is difficult to determine whether the refactor has any unexpected effect on the output.


R/compare.R
```diff
@@ -0,0 +1,274 @@
...
+##'
+##' @return An R6 object of class `orderly_packet_diff` is returned. Printing
+##'   this object will show the difference in the two packets. Additionally, the
+##'   `is_equal` method of the object return a logical indicating whether any
+##'   difference was found.
+##'
+##' @export
+orderly_compare_packets <- function(
+  target, current, search_options = NULL, root = NULL, locate = TRUE,
+  what = c("metadata", "files")) {
```
richfitz, line 107: I wonder if we should accept `verbose` here somehow? I can't immediately think of a way though and you may have already thought about this.  What I tried was:

```
> files <- orderly_compare_packets(p1, p2, what = "files", root = root)
> files
ℹ Comparing packets 20240919-155913-1eb3a847 and 20240919-155913-573ea930
! The following files exist in both packets but have different contents:
• data.R
ℹ Print the comparison with `verbose = TRUE` to display the differences in the files' contents
> files <- orderly_compare_packets(p1, p2, what = "files", root = root, verbose = TRUE)
Error in orderly_compare_packets(p1, p2, what = "files", root = root,  : 
  unused argument (verbose = TRUE)
```

richfitz: I guess it's odd to have `what` here, and not as part of the print method, given `verbose` is there.  Can we do the comparison at this point, and shift `what` there?

R/compare.R
```diff
@@ -0,0 +1,274 @@
...
+compare_filesets <- function(target, current) {
+  files <- merge(x = target, y = current, by = "path", all = TRUE)
+  status <- ifelse(
+    is.na(files$hash.x), "added",
+    ifelse(is.na(files$hash.y), "removed",
+           ifelse(files$hash.x == files$hash.y, "unchanged", "modified")))
+  data.frame(path = files$path, status = status)
+}
+
+compare_files <- function(target, current, files, root, search_options) {
```
richfitz, line 40: I feel like a non-verbose option should be really fast, and we can do this based on the hash.  That would make this sort of comparison possible even if you did not actually have the files.

It would be nice to have some option to do this quickly and just get the summary.

Similarly, there's no need to fetch the files when printing - we could just print something like "diff unavailable as files are not stored locally" or something, which would mean we could move the calls to diffobj into the print method feels more natural to me at least

richfitz (changes requested):
Thanks - this is looking great.  Some broad comments about the comparison here, which I'm happy to litigate further in a conversation before you do any more implementation if you prefer!

tests/testthat/helper-outpack.R
```diff
@@ -185,3 +185,18 @@ forcibly_truncate_file <- function(path) {
...
+#' The specified packet IDs are removed from the output, and replaced with
+#' stable strings of the same length.
+#'
+#' @param ... the list of packet IDs to remove from the output.
+#' @noRd
+scrub_packets <- function(...) {
+  ids <- c(...)
+  replacements <- sprintf("19700101-000000-%08x", seq_along(ids))
+  names(replacements) <- ids
+  function(x) stringr::str_replace_all(x, replacements)
```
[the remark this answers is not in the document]
plietar: sub/gsub only perform a single substitution, where as we need to do many. I suppose I could compose calls and make one per replacement, but that's annoying to do in a general way. This was just easier.

R/compare.R
```diff
@@ -0,0 +1,274 @@
...
+    },
+
+    format = function(verbose = FALSE, ...) {
+      target <- private$target
+      current <- private$current
+
+      cli::cli_format_method({
+        if (self$is_equal()) {
+          msg <- if (setequal(private$what, c("metadata", "files"))) {
+            "Packets {target} and {current} are identical"
```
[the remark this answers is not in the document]
plietar: "equivalent" in this case?

plietar:
diff <- orderly_compare_packets(p1, p2, root = root)
diff = list(root = root, metap1, metap2, ...)

class(diff) <- orderly_comparison

print(diff) --> summary view, no I/O
only list different field names, greyed out identical fields, ignore trivial fields (time, id, ...)

orderly_comparison_explain(diff, what # field name from metadata, verbose = FALSE)
orderly_comparison_explain()
# General purpose diff for unknown keys, we may have custom print outs for some of the keys, eg. files, dependencies, sessionInfo, ...
# verbose = TRUE might do file I/O using the saved root

S3 is.logical.orderly_comparison

plietar:
I think I’m mostly happy with the API now, except for the `verbose` argument that is a bit gross having 4 possible values.

The rendered output is a bit rough, especially when it comes to viewing the orderly specific metadata, but we can refine that over time.

richfitz (approved):
I've given this a quick whirl and I think this will be fun to try out today.  I'll let people know it's subject to change but the overall idea is now pretty solid I think. Agree that handling of `verbose` is a bit gross, and will have a think about what we can do with that

PR #148, opened 2024-06-27
Improve handling of empty or detached Git repositories.

plietar:
When orderly runs a report from a Git repository, it records information about the current state of the repository in the packet metadata. This includes the current commit hash and the name of the current branch.

It is possible for either of these to not exist, in which case we need to make sure orderly behaves gracefully. The commit hash will be missing on a brand new repository which hasn't had any commits yet (or equivalently, an existing repository where `git checkout --orphan` was used. The branch name would be missing if the repository is in a "detached HEAD" state, in other words the user checked out a specific commit hash instead of a named branch.

The first situation, of an empty repository, was not supported by orderly at all. It would throw an error and fail to run the report entirely. In the second case, when on a detached HEAD, the Git library we use reports the branch name as "HEAD", and that is what we were recording in the metadata. This is a bit misleading since it is not an actual branch name.

Both cases have been modified to work without errors and to omit the missing values from the metadata.

### DIFF
--- a/R/outpack_misc.R
+++ b/R/outpack_misc.R
@@ -44,8 +44,16 @@ git_info <- function(path) {
   if (is.null(repo)) {
     return(NULL)
   }
-  list(sha = gert::git_commit_id(repo = repo),
-       branch = gert::git_branch(repo = repo),
+
+  branch <- gert::git_branch(repo = repo)
+  if (identical(branch, "HEAD")) {
+    # HEAD isn't a valid branch name, and instead is what gets returned when a
+    # detached head was checked out.
+    branch <- NULL
+  }
+
+  list(sha = ignore_errors(gert::git_commit_id(repo = repo)),
+       branch = branch,
        url = gert::git_remote_list(repo = repo)$url)
 }
 
--- a/R/outpack_misc.R
+++ b/R/outpack_misc.R
@@ -45,16 +45,18 @@ git_info <- function(path) {
     return(NULL)
   }
 
+  sha <- tryCatch(gert::git_commit_id(repo = repo),
+                  error = function(e) NA)
+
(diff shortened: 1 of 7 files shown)


plietar:
I'm kind of on the fence whether the fields in the JSON object should be omitted or set to null. I went with the latter for now, but I'm open to change that. Once we agree on what to do I will need to make a change to the upstream schema first before merging this PR.

richfitz:
> I'm kind of on the fence whether the fields in the JSON object should be omitted or set to null

Set to `null` I think

R/outpack_misc.R
```diff
@@ -44,9 +44,19 @@ git_info <- function(path) {
...
-  list(sha = gert::git_commit_id(repo = repo),
-       branch = gert::git_branch(repo = repo),
-       url = gert::git_remote_list(repo = repo)$url)
+
+  sha <- tryCatch(gert::git_commit_id(repo = repo),
+                  error = function(e) NA)
+
+  branch <- gert::git_branch(repo = repo)
+  if (is.null(branch) || identical(branch, "HEAD")) {
+    # NULL can be returned when working in a repo that has no commits yet.
```
richfitz, line 53: I wonder if it's worth having sentinel values for these - could be `(empty)` for no commits and `(detached)` for detached head, or tracking this though something in the metadata? Probably not worth getting too wound up about though

inst/schema/outpack/git.json
```diff
@@ -7,11 +7,11 @@
     "type": "object",
     "properties": {
         "sha": {
-            "type": "string",
+            "type": ["string", "null"],
```
richfitz, line 10: This is my preference for the schema change

richfitz (approved):

PR #160, opened 2024-07-05
Harmonise orderly_artefact() with orderly_dependency etc

richfitz:
This PR does part 1 of harmonising `orderly_artefact()` with `orderly_resource` and friends:

* the `description` argument, which other functions lack becomes optional in the schema and the function definition
* warn if orderly_dependency() is used where the description arg is unnamed

This latter change, once in the wild, will make it less disruptive to change the order of the arguments, so that we have

```
orderly_artefact <- function(files, description = NULL) ...
```


plietar (commented):
Is there any part of Packit that uses the artefacts from the metadata to view packets?

plietar (approved):

PR #162, opened 2024-07-10
Add scripts and workflows to manage version number

plietar:
These were copied almost verbatim from hipercow. The only difference is that I removed the bits of code related to keeping hipercow.windows in sync.


richfitz (approved):

richfitz:
thanks, definitely at the point where we want this now

PR #163, opened 2024-07-15
Optional verbose output for orderly_location_list

richfitz:
This PR adds support for returning verbose information about which locations are configured.  This means that the return type of the function varies on an argument which is not great, but I think it's nicer than two functions.  For the 'verbose' output we already hold a data.frame with exactly the information that might be wanted so we return this.  There's some noise as all calls to `orerly_location_list(root)` without naming the `root` argument failed

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -203,18 +203,31 @@ orderly_location_remove <- function(name, root = NULL, locate = TRUE) {
 ##'
 ##' @inheritParams orderly_metadata
 ##'
-##' @return A character vector of location names. The special name
-##'   `local` will always be present.
+##' @param verbose Logical, indicating if we should return a
+##'   data.frame that includes more information about the location.
+##'
+##' @return Depending on the value of `verbose`:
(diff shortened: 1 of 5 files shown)


R/location.R
```diff
@@ -203,18 +203,31 @@ orderly_location_remove <- function(name, root = NULL, locate = TRUE) {
...
 ##'
-##' @return A character vector of location names. The special name
-##'   `local` will always be present.
+##' @param verbose Logical, indicating if we should return a
+##'   data.frame that includes more information about the location.
+##'
+##' @return Depending on the value of `verbose`:
+##'
+##' * `verbose = FALSE`: A character vector of location names. The
+##'   special name `local` will always be present.  This is the
```
plietar, line 212: This line about local should be moved out of the bullet points into the general description.

As it stands it seems to imply that with `verbose=TRUE` the return value does not include local.

plietar (approved):

PR #165, opened 2024-07-26
Replace uses of `fs::file_copy` with `file.copy`.

plietar:
Unfortunately `fs::file_copy` has issues when working with read-only files on a Samba mounted file system. `file.copy` doesn't have the same problem. See mrc-5557 for more details.

This was tested manually and confirmed to work by running a basic report with an `orderly_dependency` call on a Samba mount. Unfortunately it is not something we can easily test for automatically.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.24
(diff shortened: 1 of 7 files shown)


richfitz (approved):
There's one use left of fs::file_copy in the tests - that's fine as we never expect that to be used on a network share

PR #166, opened 2024-08-22
Improve handling of non-JSON errors.

plietar:
When getting a non-200 HTTP response, orderly tries to decode the body as JSON in order to extract the error message. We do however occasionally get errors that aren't JSON formatted, in particular if the error is generated by nginx rather than Packit.

This situation can arise if the Packit server is down (a 503 error), or if nginx rejects the request due to the body being too large (a 413 error).

We can use the content type header of the response to determine whether it is safe to decode the body as JSON. If the response is not of the expected mime-type, the body is ignored and a generic error based on the status code alone is returned.

### DIFF
--- a/tests/testthat/helper-outpack-http.R
+++ b/tests/testthat/helper-outpack-http.R
@@ -2,7 +2,7 @@ mock_headers <- function(...) {
   structure(list(...), class = c("insensitive", "list"))
 }
 
-mock_response <- function(content, status = 200, wrap = TRUE, download = NULL) {
+mock_response <- function(content, status = 200L, wrap = TRUE, download = NULL) {
   headers <- mock_headers()
   if (!is.null(download)) {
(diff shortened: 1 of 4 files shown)


tests/testthat/helper-outpack-http.R
```diff
@@ -2,7 +2,7 @@ mock_headers <- function(...) {
   structure(list(...), class = c("insensitive", "list"))
 }
 
-mock_response <- function(content, status = 200, wrap = TRUE, download = NULL) {
+mock_response <- function(content, status = 200L, wrap = TRUE, download = NULL) {
```
richfitz, line 5:
```suggestion
mock_response <- function(content, status = 200L, wrap = TRUE, 
                          download = NULL) {
```

this will keep the linter happy

richfitz (approved):
one inline suggestion to appease the linter, otherwise LGTM

PR #168, opened 2024-08-30
Implement OAuth device flow for GitHub logins.

plietar:
GitHub-based logins to Packit rely on users specifying a personal access token when adding the location. This has a number of usability and security disadvantages.

Users need to manually take steps in order to create a PAT, and they need to make sure they have selected the right scope. It is quite likely users would be tempted to create tokens with way more scopes than is necessary.

The configured PAT would be stored in the orderly configuration, which many users share and keep on network drives that are accessible to everyone.

By using an OAuth flow to generate our own tokens, we remove the extra steps the user needed to take and we can make sure it has just the right scopes needed. Additionally the token cache implemented by httr2 is located outside of the orderly repository, in the user's home directory.

The device flow works by showing the user an 8 character code in the console, and a link to https://github.com/login/device. The user needs to type in the code and approve the login. Meanwhile, the orderly2 client polls the GitHub API until the authentication has been approved.

There are other OAuth flows we could have used, in particular the authorization code flow (with PKCE). In this case, a link is printed to the console and the user needs to follow it and approve the login. They are then redirected to a localhost server, which has been started by orderly2. This removes the need to need with the 8 character code of the device flow, but it is a bit finicky and assumes orderly2 is running on the same machine as the user's browser, which may not be the case when runing over SSH. For this reason, the device flow is preferred.

It is still possible to specify a PAT as before, in which case the OAuth flow is skipped. This can be useful in non-interactive situations, such as a CI pipeline.

Note that there are still fundamental issues with this model. In particular, the Github tokens are not bound to the application. This means that a malicious application could manage to get a user to sign in to it, obtain a access token for the user, and use this to login to Packit. From just a GitHub `read:org` scope, the app is able to escalate to full impersonation of the user on the Packit server.

Additionally it still tightly couples the orderly2 client with GitHub as an authentication method, prohibiting the use of other authentication methods, such as basic username / password.

Ideally we would remove the use of Github Access tokens as a way to login to Packit. Instead, Packit should itself implement an OAuth2 Authorization Server. orderly2 would perform the device flow directly against Packit, and directly retrieve a Packit JWT in exchange. During the device flow, when the user is directed to the Packit website to enter the user code, Packit may require the user to login. When this happens, a new, parallel OAuth flow would take place, this time with GitHub as the OAuth AS and Packit as the OAuth client (using the authorization code flow). When the nested flow is complete and login is succesful, the user would be prompted to type in the code and complete the flow initiated by orderly2.

With this approach, orderly2 has no involvement with GitHub at all, and Packit can use another authentication method (or even no authentication). This would be completely transparent to the orderly2 client. It also provides tighter binding between GitHub and Packit, as Packit fetches the access token directly from the former, and cannot be supplied a token from an arbitrary application.

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -41,8 +41,8 @@
 ##' * `url`: The location of the server
 ##'
 ##' * `token`: The value for your your login token (currently this is
-##'   a GitHub token with `read:org` scope).  Later we'll expand this
-##'   as other authentication modes are supported.
+##'   a GitHub token with `read:org` scope). If missing or NULL, orderly2 will
+##'   perform an interactive authentication against GitHub to obtain one.
 ##'
 ##' **Custom locations**:
 ##'
@@ -110,6 +110,7 @@ orderly_location_add <- function(name, type, args, root = NULL, locate = TRUE) {
     assert_scalar_character(loc$args[[1]]$url, name = "args$url",
                             call = environment())
     assert_scalar_character(loc$args[[1]]$token, name = "args$token",
+                            allow_null = TRUE,
                             call = environment())
   }
 
@@ -764,7 +765,7 @@ new_location_entry <- function(name, type, args, call = NULL) {
   } else if (type == "http") {
     required <- "url"
   } else if (type == "packit") {
-    required <- c("url", "token")
+    required <- c("url")
   } else if (type == "custom") {
     required <- "driver"
   }
--- a/R/location.R
+++ b/R/location.R
@@ -765,7 +765,7 @@ new_location_entry <- function(name, type, args, call = NULL) {
   } else if (type == "http") {
     required <- "url"
   } else if (type == "packit") {
-    required <- c("url")
+    required <- "url"
   } else if (type == "custom") {
     required <- "driver"
   }
--- a/R/location.R
+++ b/R/location.R
@@ -44,6 +44,10 @@
 ##'   a GitHub token with `read:org` scope). If missing or NULL, orderly2 will
 ##'   perform an interactive authentication against GitHub to obtain one.
 ##'
+##' * `save_token`: If no token is provided and interactive authentication is
+##'    used, this controls whether the GitHub token should be saved to disk.
+##'    Defaults to TRUE if missing.
+##'
 ##' **Custom locations**:
 ##'
 ##' All outpack implementations are expected to support path and http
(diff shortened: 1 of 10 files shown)


plietar:
Tests are broken on R-devel, see https://github.com/mrc-ide/orderly2/pull/169

R/location_packit.R
```diff
@@ -0,0 +1,75 @@
...
+  # used by Packit. It should be fine to hardcode a value regardless of which
+  # server we are talking to.
+  httr2::oauth_client(
+    id = "Ov23liUrbkR0qUtAO1zu",
+    token_url = "https://github.com/login/oauth/access_token",
+    name = "orderly2"
+  )
+}
+
+do_oauth_device_flow <- function(base_url) {
```
richfitz, line 12: It's unfortunately a httr2 thing, but I was confused the first time seeing the prompt:

```
ℹ Logging in to https://packit.dide.ic.ac.uk/malariaverse-sitefiles/
→ Copy 18C0-2805 and paste when requested by the browser
Press <enter> to proceed:
```

At this point I thought that I was waiting for a browser window to open, then I tried clicking the link. Finally I pressed enter and it launched the browser.  It might be worth here (or around l39) posting a small hint about how the login process will go?

richfitz (approved):
working for me with both oauth and tokens - some comments included

R/location_packit.R
plietar: Yeah this is quite annoying. It also doesn't work if R can't open a browser, eg. over SSH.
I'll fix something.

plietar: Alright so if I override interactive temporarily, the behaviour is a lot less confusing. I've changed it, let me know if this helps.

R/location.R
```diff
@@ -41,8 +41,8 @@
 ##' * `url`: The location of the server
 ##'
 ##' * `token`: The value for your your login token (currently this is
-##'   a GitHub token with `read:org` scope).  Later we'll expand this
-##'   as other authentication modes are supported.
+##'   a GitHub token with `read:org` scope). If missing or NULL, orderly2 will
+##'   perform an interactive authentication against GitHub to obtain one.
```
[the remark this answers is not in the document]
plietar: Done, added a `save_token` argument.

richfitz (approved):

PR #170, opened 2024-09-02
Re-enable some integration tests on windows.

plietar:
These were disabled when they were introduced, back in mrc-ide/orderly2#42. They seem to work now, possibly thanks to mrc-ide/outpack_server#48, so no need to skip them anymore. We run CI on outpack_server on all platforms these days, so it should be fairly stable.


richfitz:
I presume this will pass on R-devel once main is merged in, now that #169 is merged?

richfitz (approved):
Thanks!

PR #171, opened 2024-09-02
Improve handling of deleted implicit inputs.

plietar:
In implicit mode, orderly prints a message if some of the implicit inputs have been modified by the report. However the code to enumerate modified files did not properly handle files that have been deleted. When that happens, `fs::file_info` return NA in all the metadata columns.

We now detect these files and display a slightly different warning for it.

Cleaned up a bit of adjacent code: switch from handcrafted bullet points to using `cli::cli_ul`, fixed some tests which had used `suppressMessages` when they could have used `orderly_run_quietly`.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.30
+Version: 1.99.31
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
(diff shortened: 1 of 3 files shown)


plietar:
I'm kind of on the fence about whether this even deserves a warning at all. There's nothing really actionable the user can do other than not writing weird reports.

richfitz (approved):
Thanks, and for getting this found and fixed so quickly.

> I'm kind of on the fence about whether this even deserves a warning at all. There's nothing really actionable the user can do other than not writing weird reports.

The intention had been to expose something like `outpack_packet_file_mark` to mark files as ignored; we should expose some mechanism for doing this and hint at it in the warning.  But I think do that in a subsequent PR

PR #172, opened 2024-09-03
Cache builds of outpack_server

plietar:
Compiling outpack_server each time takes about 2 minutes on Linux and up to 5 minutes on Windows, slowing down our CI jobs. We can cache the resulting binaries to avoid having to do this on every orderly2 build.

The action includes the Git revision of the outpack_server repository in its cache key, meaning we'll automatically pull new versions of the server as they come. Of course we'll need to compile these, but we will only be paying the cost once per new commit, instead of each time. The development rate on outpack_server is low enough that this shouldn't be too cumbersome.

The alternative would be to start tagging releases of outpack_server and publish pre-built binaries of it. It is more effort, both to setup and to use in the future, so simply caching the installation here seems preferable.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -2,9 +2,9 @@
 # Need help debugging build failures? Start at https://github.com/r-lib/actions#where-to-find-help
 on:
   push:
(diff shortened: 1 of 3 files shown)


M-Kusumgar (approved):
LGTM!

PR #175, opened 2024-09-04
Propagate name parameter in match_value.

plietar:
Without this, when the first argument of `match_value` was not a valid character scalar, the function would unconditionally call it `x`, even if the caller's code makes no mention of `x`.

```r
match_value(5, list("foo", "bar"))
#> Error:
#> ! 'x' must be character
```

With this change, the name as used in the caller of `match_value` is used, which gives a less confusing output:

```r
match_value(5, list("foo", "bar"))
#> Error:
#> ! '5' must be character
```

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.32
+Version: 1.99.33
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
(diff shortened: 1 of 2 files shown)


plietar:
@richfitz do you know what the `arg` that gets passed around throughout the file is for? It just gets passed to `cli::cli_abort` but I don't see any mention of it in the `cli` docs.

richfitz (approved):

richfitz:
See `?rlang::abort`, where it is used - in the error it will highlight the offending argument (I think you will see that with `rlang::last_trace` or `rlang::last_error`.

I've tidied a few of these things up as https://github.com/reside-ic/reside.utils/ which we can pull in for reuse in orderly among others, I'll get a PR together for this perhaps on the train on the way home tonight

PR #177, opened 2024-09-05
Handle error responses without any content type.

plietar:
In some circumstances, Packit returns an HTTP error with no body and no Content-Type header. In those cases, httr2 returns NA as the content type.

#166 added a check to make sure the content type is JSON before decoding the error, however the check did not handle NA values properly and gave a `missing value where TRUE/FALSE needed`.

This now uses `identical` instead of `==` to compare the content type string against the expected string.


richfitz (approved):
Thanks, but what a nuisance!

PR #178, opened 2024-09-10
Use the fs package to copy files.

plietar:
It has more bells and whistles, and is often more performant. In particular, it makes use of `CopyFileW` (on Windows) and `copy_file_range` (on Linux) when applicable. This offers a big speedup when copying a file within a network drive, which is a common use case for orderly.

The reason why `file.copy` was introduced in #165 is that copying read-only files with the `fs` package wasn't working properly on Samba drives. It turns out this situation only occurs when `overwrite = TRUE`. When the flag is set to FALSE, the copy succeeds. To work around this we do an initial check before the copy to delete any files in the destination path, and copy with `overwrite = FALSE`. This was manually tested by running the test suite with the `TMPDIR` environment variable set to a path on a network drive.

### DIFF
--- a/R/outpack_store.R
+++ b/R/outpack_store.R
@@ -33,9 +33,14 @@ file_store <- R6::R6Class(
         stop(not_found_error(message, missing))
       }
 
-      # Files in the archive are read-only. It's easier on the user if we make
-      # them writable again.
-      copy_files(src, dst, overwrite = overwrite, make_writable = FALSE)
+      copy_files(src, dst, overwrite = overwrite)
+
+      # Files in the store are read-only to avoid accidental corruption.
+      # This is however an implementation detail, and we should export them as
+      # writable again.
+      if (length(dst) > 0) { # https://github.com/r-lib/fs/issues/471
+        fs::file_chmod(dst, "u+w")
+      }
 
       invisible(dst)
     },
--- a/R/outpack_store.R
+++ b/R/outpack_store.R
@@ -32,11 +32,10 @@ file_store <- R6::R6Class(
                            paste(sprintf("  - %s", missing), collapse = "\n"))
         stop(not_found_error(message, missing))
       }
-      fs::dir_create(dirname(dst))
 
-      # Set copy_mode = FALSE: files in the store are read-only. It's easier on
-      # the user if we make them writable again.
-      copy_files(src, dst, overwrite = overwrite, copy_mode = FALSE)
+      # Files in the archive are read-only. It's easier on the user if we make
+      # them writable again.
+      copy_files(src, dst, overwrite = overwrite, make_writable = FALSE)
 
       invisible(dst)
     },
--- a/R/run.R
+++ b/R/run.R
@@ -194,7 +194,8 @@ orderly_run <- function(name, parameters = NULL, envir = NULL, echo = TRUE,
 
   if (dat$strict$enabled) {
     inputs_info <- NULL
-    copy_files(file.path(src, entrypoint_filename), path)
+    copy_files(file.path(src, entrypoint_filename),
+               file.path(path, entrypoint_filename))
   } else {
     inputs_info <- copy_resources_implicit(src, path, dat$resources,
                                            dat$artefacts)
@@ -477,8 +478,7 @@ copy_resources_implicit <- function(src, dst, resources, artefacts) {
   }
 
   copy_files(file.path(src, to_copy),
-             file.path(dst, to_copy),
-             overwrite = TRUE)
+             file.path(dst, to_copy))
 
   withr::with_dir(dst, fs::file_info(to_copy))
 }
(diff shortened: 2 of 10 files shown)


tests/testthat/helper-outpack.R
```diff
@@ -184,7 +184,7 @@ outpack_packet_end_quietly <- function(...) {
 
 forcibly_truncate_file <- function(path) {
   permissions <- fs::file_info(path)$permissions
-  fs::file_chmod(path, "a+w")
-  file.create(path)
+  fs::file_delete(path)
+  fs::file_create(path)
```
plietar, line 188: Somehow, chmod isn't immediately visible when working over Samba, even after it returns.
I had to do this to work around it and get the tests passing.

```sh
touch hello.txt
chmod 444 hello.txt; chmod 666 hello.txt; ls -l hello.txt
```

Running the second line many times, it will sometimes print the permissions as `-rwxr-xr-x`, sometimes as `-r-xr-xr-x`.

R/run.R
```diff
@@ -477,8 +478,7 @@ copy_resources_implicit <- function(src, dst, resources, artefacts) {
   }
 
   copy_files(file.path(src, to_copy),
-             file.path(dst, to_copy),
-             overwrite = TRUE)
+             file.path(dst, to_copy))
```
plietar, line 481: This function only ever gets called with a completely empty `dst` directory. There is no reason to use `overwrite = TRUE`.

plietar:
The tests all pass on Linux with a `TMPDIR` set to a network drive.

Depending on the libcurl version it may need the latest httr2 to fix https://github.com/r-lib/httr2/issues/534
The leaked file handles prevent the test cleanup from removing some directories.

R/util.R
```diff
@@ -152,16 +152,44 @@ expand_dirs <- function(paths, workdir) {
...
+  #
+  # If you are going to make changes to this function, make sure to run all
+  # tests with TMPDIR set to a path on a network drive.
+  if (overwrite) {
+    exists <- fs::file_exists(dst)
+    fs::file_delete(dst[exists])
+  }
+
+  fs::file_copy(src, dst, overwrite = FALSE)
+  if (make_writable && length(dst) > 0) {
```
richfitz, line 191: do you want to move this length(dst)> 0 check to be an early exit at the point you check that src and dst are the same length?

R/outpack_store.R
```diff
@@ -32,11 +32,10 @@ file_store <- R6::R6Class(
                            paste(sprintf("  - %s", missing), collapse = "\n"))
         stop(not_found_error(message, missing))
       }
-      fs::dir_create(dirname(dst))
 
-      # Set copy_mode = FALSE: files in the store are read-only. It's easier on
-      # the user if we make them writable again.
-      copy_files(src, dst, overwrite = overwrite, copy_mode = FALSE)
+      # Files in the archive are read-only. It's easier on the user if we make
+      # them writable again.
```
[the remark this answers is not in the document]
plietar: Huh my mistake here is actually setting `make_writable = FALSE` instead of `TRUE`. I'm amazed there's not a test to catch that. I'll fix that, also need to change `archive` to `store` in the comment.

But otherwise the comment is correct. This function moves files out of the store and into the user's directory (either the drafts folder if running a packet, the src folder if running interactively or anywhere if the user just called `orderly_copy_files`).

R/util.R
plietar: Honestly I would rather remove it entirely, but I can't because `fs` doesn't handle it correctly: https://github.com/r-lib/fs/issues/471 Keeping it to as small of a scope is the best compromise in my opinion, until that issue is fixed.

The rest of the function works the same for `n = 0`, and I see no reason to introduce special cases.

R/run.R
plietar: I'm sorry I don't really understand the comment. The only "previous call" in this file also omits the `make_writable` argument. `make_writable = FALSE` is a no-op, it is not going to make things read-only (not that we would want that here anyway).

Maybe having `copy_files` do the chmod is confusing, and instead I remove the `make_writable` argument and move the chmod to the only two places it is actually needed.

richfitz: previous call was outpack_store.R:38 where you have `make_writable = FALSE`, but it looks like that should be TRUE :)

R/outpack_store.R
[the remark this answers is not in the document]
plietar: Hah, so the tests passed because it was doing `orderly_copy_files` twice and whether that works. Turns out that now works even if the file is read-only because the new `overwrite = TRUE` behaviour will overwrite read-only files, whereas the old one did not.

I think I might actually tweak the `copy_files` function to not allow that anymore.

plietar:
Okay so the last commit actually removes the `make_writable` argument. I don't really think there's a good reason to have `copy_files` do all the things, and it's just a `fs::file_chmod` call for the caller.

Also now refuses to overwrite read-only files, to match the default `fs::file_copy(overwrite = TRUE)` et al behaviour. The implementation is obviously very racy.

richfitz (approved):

PR #179, opened 2024-09-11
Refactor directory expansion.

plietar:
There were three separate implementations of directory expansion spread across the codebase, each accepting slightly different inputs. The one used by `orderly_resource` operated over a single vector of paths.  The one used by `orderly_shared_resource` operated over a dataframe of `there`/`here` columns. Finally the one used by `orderly_copy_files` used the packet metadata to enumerate files, instead of looking at the filesystem.

The three implementations are replaced by just one, which uses callbacks to determine which files are directories and to enumerate their contents. This enables it to be used either with real files on disk or off of a packet's metadata.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.36
+Version: 1.99.37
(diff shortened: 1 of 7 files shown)


richfitz (approved):
Fabulous, thanks for harmonising that

PR #181, opened 2024-09-20
Add support for specifying Packit tokens directly.

plietar:
As part of experimenting with new authentication methods for Packit, it is useful to be able to provide Packit tokens directly to orderly2. Until now, orderly2 would only accept GitHub tokens, and would perform a request to exchange it for a Packit token.

Thankfully, GitHub tokens have a [well defined and documented prefix][ghblog], of the form `ghp_`, `gho_`, ... The underscore makes sure it so that it never could never be confused with a JWT. Thanks to this, the distinction between Packit and GitHub tokens is unambiguous.

Eventually we may want to implement some of these authentication methods in orderly2 itself, but it the meantime this makes experimentation easier.

[ghblog]: https://github.blog/engineering/platform-security/behind-githubs-new-authentication-token-formats/

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.39
+Version: 1.99.40
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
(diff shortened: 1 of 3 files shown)


plietar:
For an example of this being used https://github.com/plietar/orderly-ci-test/blob/f1987986d5fceeaf7b4cceb9aacf2708ae776f0a/run.R

Eventually we'll probably want to export that code somewhere, either in orderly2 or in a reusable action.

richfitz (approved):

PR #183, opened 2024-10-17
Drop `locate` argument everywhere public-facing

richfitz:
This PR drops the `locate` argument which was inconsistently present, had inconsistent arguments and was inconsistently used.  We now search for the root from `root` (usually `NULL`) in all cases.

Calls to `root_open` have been slightly simplified by adding a reasonable default for `call`; doing that everywhere is another bit of maintenance we can do later

### DIFF
--- a/R/root.R
+++ b/R/root.R
@@ -140,7 +140,8 @@ empty_config_contents <- function() {
 ## * also check that the outpack and orderly path are compatibible
 ##   (this is actually quite hard to get right, but should be done
 ##   before anything is created I think)
-root_open <- function(path, locate, require_orderly = FALSE, call = NULL) {
+root_open <- function(path, locate = TRUE, require_orderly = FALSE,
+                      call = parent.frame()) {
   if (is.null(call)) {
     call <- environment()
   }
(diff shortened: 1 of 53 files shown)


R/root.R
```diff
@@ -140,7 +139,7 @@ empty_config_contents <- function() {
 ## * also check that the outpack and orderly path are compatibible
 ##   (this is actually quite hard to get right, but should be done
 ##   before anything is created I think)
-root_open <- function(path, locate, require_orderly = FALSE, call = NULL) {
+root_open <- function(path, require_orderly = FALSE, call = parent.frame()) {
   if (is.null(call)) {
```
plietar, line 143: Can we remove this line? Presumably no where calls this with an explicit `call = NULL`.

R/root.R
```diff
@@ -140,7 +139,7 @@ empty_config_contents <- function() {
 ## * also check that the outpack and orderly path are compatibible
 ##   (this is actually quite hard to get right, but should be done
 ##   before anything is created I think)
-root_open <- function(path, locate, require_orderly = FALSE, call = NULL) {
+root_open <- function(path, require_orderly = FALSE, call = parent.frame()) {
```
plietar, line 142: Seems you've added values for `require_orderly` everywhere - would it be worth removing the default?

plietar (approved):
LGTM, just two minor further possible cleanups.

PR #184, opened 2024-10-17
Add user-friendly functions to add locations

richfitz:
This annoyed me every time I tried to show someone orderly, so fixing it now before I teach it to a room of people.

I was trying to be too general with `orderly_location_add` but we can make it much nicer if we have functions `orderly_location_add_path` etc with nice arguments, docs and autocomplete for them.

For custom locations we might add similar functions within their packages then if users use `library()` they'll be available.

A very noisy diff but this is mostly just using the new version throughout the tests


R/location.R
```diff
@@ -3,9 +3,12 @@
...
 ##' locations. Note that adding a location does *not* pull metadata
 ##' from it, you need to call
-##' [orderly2::orderly_location_pull_metadata] first.
+##' [orderly2::orderly_location_pull_metadata] first.  The function
+##' `orderly_location_add` can add any sort of location, but the other
+##' functions documented here (`orderly_location_add_path`, etc) will
+##' typically be much easier to use in practice.
 ##'
-##' We currently support two types of locations - `path`, which points
+##' We currently support three types of locations - `path`, which points
```
plietar, line 11: two types of locations and off by one errors :grin:

R/location.R
```diff
@@ -132,6 +116,44 @@ orderly_location_add <- function(name, type, args, root = NULL, locate = TRUE) {
 }
 
 
+##' @rdname orderly_location_add
+##'
+##' @param path The path to the other archive root. This should
+##'   generally be an absolute path, or the behaviour of outpack will
+##'   be unreliable.
```
plietar, line 123: Might be worth cleaning up and defining the semantics of a relative path at some point (not in this PR). IMO it should be relative to the orderly root, and be resolved each time we use the location, not when it is first added.

R/location.R
```diff
@@ -132,6 +116,44 @@ orderly_location_add <- function(name, type, args, root = NULL, locate = TRUE) {
...
+  orderly_location_add(name, "http", args, root = root)
+}
+
+
+##' @rdname orderly_location_add
+##'
+##' @param token The value for your your login token (currently this
+##'   is a GitHub token with `read:org` scope). If `NULL`, orderly2
+##'   will perform an interactive authentication against GitHub to
+##'   obtain one.
```
plietar, line 145: Technically this can be a Packit JWT instead of a GitHub PAT, but we don't really publicize any way of obtaining those so I think it is fine if you don't want to mention it here.

This is the only place a JWT would be used today: https://github.com/mrc-ide/orderly-action/blob/main/packit-login.R#L65-L69

plietar (approved):
Yes thank you. This has annoyed me many times before as well.

One alternative would be to have a generic orderly_location_add, which just takes a string and figures it out, like Git does it. We would need some kind of negotiation to determine if an http URL is an outpack_server or a Packit instance (or we use `packit+https://`, but that's ugly). We would lose some flexibility if we did that though.

In any case this PR is a good step forward already as is.

As a side-note, at some point maybe before next week's workshop I want to get Packit to display on its home page the line of code needed to add the location to a local workspace, in an easily copy-pasteable way, similar to how GitHub does for new repositories.

R/location.R
richfitz: Are you thinking here again of people on the cluster where directories "move"?

R/location.R
richfitz: I just copied this in from the string that was present in the Details for now

richfitz:
> One alternative would be to have a generic orderly_location_add, which just takes a string and figures it out, like Git does it. We would need some kind of negotiation to determine if an http URL is an outpack_server or a Packit instance (or we use packit+https://, but that's ugly). We would lose some flexibility if we did that though.

No objection to supporting that; it's concice and would be portable to the Python version too.  But it would make custom locations hard to support, and we already have a couple of those now.

> As a side-note, at some point maybe before next week's workshop I want to get Packit to display on its home page the line of code needed to add the location to a local workspace, in an easily copy-pasteable way, similar to how GitHub does for new repositories.

This is a great idea

PR #185, opened 2024-10-18
Make pulling metadata more chatty

richfitz:
This PR makes pulling metadata a bit more chatty, which removes the feeling that it has done nothing when pulling from path locations; if I find that alarming I imagine that users will too.  I toyed with cli's "simple progress bar" api but instead just went with printing a few messages, which feels reasonable.

After our discussion about having to suppress the output of this everywhere, I have taken the liberty of adding an option `orderly.quiet` which I have set in the tests and put this behind.  We use this pattern elsewhere at least, for example: https://mrc-ide.github.io/hipercow/articles/details.html#hipercow-progress - this means that the fallout in the tests is minimal and I think the extra logic not that bad to think about.

Somehow in two package options we were already inconsistent so I've updated `orderly_index_progress` to `orderly.index_progress`

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.46
+Version: 1.99.47
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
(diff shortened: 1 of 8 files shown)


plietar:
Is the intention that `if (!quiet)` would eventually be added to large portions of the messages we emit? If so, having quiet as a function parameter seems a bit superfluous, as we'll end up having to add it to literally almost every function. Having the global `orderly.quiet` option seems sufficient to me.

I'm not sure I like setting the option in `tests/testthat/helper-orderly.R`, given that it ends up being used if I do `devtools::load_all()` (which is 90% of how I load orderly).

Instead of setting the option, could we use the output of [`testthat::is_testing()`](https://testthat.r-lib.org/reference/is_testing.html) as the fallback if no option is set?

plietar (approved):

Issue #156, opened 2024-07-04
orderly_metadata_extract(options = list(allow_remote = TRUE, location = "root2"))


jeffeaton:
* If I explicitly specify a remote location, requiring me to also specify `allow_remote = TRUE` feels redundant and error prone:

```
> ## Oops this doesn't give me accurate information about root1!
> orderly_metadata_extract(options = list(location = "root1"))
[1] id         name       parameters
<0 rows> (or 0-length row.names)

> ## I have to specify allow_remote = TRUE to see what's there
> orderly_metadata_extract(options = list(location = "root1", allow_remote = TRUE))
                        id  name parameters
1 20240704-215806-90fbb966 hello            

> ## **After** I pull the packet, it shows me the info about root1 without allow_remote
> orderly_location_pull_packet(name = "hello")
ℹ Looking for suitable files already on disk
ℹ Need to fetch 2 files (112 B) from 1 location
✔ Fetched 2 files (112 B) from 'root1' in 36ms.         

> orderly_metadata_extract(options = list(location = "root1"))
                        id  name parameters
1 20240704-215806-90fbb966 hello           
```

* Personally, preference for `allow_remote = TRUE` to be default. For me an interface that seems intuitive is:
   *  Always show what is in all of the locations
   * If I want only the local location, specify this as `orderly_metadata_extract(options = list(location = "local"))`
   * Add a column to the table returned that indicates whether packet is (fully) in the local location or not—which I guess is what specifying `allow_remote` is intending to control?

* Would prefer for `location = ` to be a named argument: `orderly_metadata_extract(location = "root1")`

Issue #154, opened 2024-07-04
orderly_location_pull_packet() doesn't require location= specification


jeffeaton:
Collaborators were surprised and slightly confused today that `orderly_location_pull_packet()` did not require specification of a location from which pulling.

That also gave rise to a similar question of what happens if the same packet exists at multiple locations—which is it pulling from?

Is the answer that it doesn't matter because the content must be the same?  Will it default to pulling from whichever location appears first in `orderly_location_list()`?

I notice in the message below that message specifies it came from `root1`:
```
> orderly_location_pull_packet(name = "hello")
ℹ Looking for suitable files already on disk
ℹ Need to fetch 2 files (112 B) from 1 location
✔ Fetched 2 files (112 B) from 'root1' in 41ms.
```
-----
Here's a little example I used to explore the behaviour for myself:
```
library(orderly2)

root1 <- tempfile()
root2 <- tempfile()
root3 <- tempfile()

## Create a report at root1
orderly_init(root1)

setwd(root1)
orderly_new("hello")

writeLines('
library(orderly2)
orderly_artefact("hello", "hello.txt")
writeLines("hello world!", "hello.txt")
',
"src/hello/hello.R"
)

id <- orderly_run("hello")


## Pull packet to root2
orderly_init(root2)
setwd(root2)

orderly_location_add("root1", "path", list(path = root1))
orderly_location_pull_packet(name = "hello", options = list(pull_metadata = TRUE))


## Create root3; add both root1 and root2 as locations
orderly_init(root3)
setwd(root3)

orderly_location_add("root1", "path", list(path = root1))
orderly_location_add("root2", "path", list(path = root2))

orderly_metadata_extract(options = list(allow_remote = TRUE, pull_metadata = TRUE))

orderly_metadata_extract(options = list(allow_remote = TRUE, location = "root1"))
orderly_metadata_extract(options = list(allow_remote = TRUE, location = "root2"))

orderly_location_pull_packet(name = "hello")
```

richfitz:
See https://mrc-ide.github.io/orderly2/reference/orderly_search_options.html for details - this is behaving as expected and as designed, and is perhaps something that you will also have to unlearn from orderly1.

You have a series of remotes - most users will have zero or one.  When pulling the default is to pull from **all** remotes; that is the query evaluated over the graph of packets contained everywhere.  You can restrict this by using the `location` argument to the options there, as you have seen.

The message

```
[...]

PR #186, opened 2024-10-18
Ergonomic improvements for querying

richfitz:
As discussed by Jeff in #156 and elsewhere (#154, offline), it is hard to know what to put into functions that might interact with metadata on remotes.  This PR starts the process of improving this.

* If the user specifies a location (non-`NULL` argument `location`), then `allow_remote` should default to `TRUE`.  I've implemented this by changing the advertised default to `NULL` and evaluating it after seeing what we were given for `location`.
* The arguments to compose a query (`expr`, `name`, etc) would be useful on user-facing functions for autocomplete etc.  Otherwise they are several levels of man page away and really quite hard to find

The improvements continue in the more disruptive PR #187

### DIFF
--- a/R/query_search.R
+++ b/R/query_search.R
@@ -24,6 +24,7 @@
 ##'   options are used (i.e., `orderly2::orderly_search_options()`)
 ##'
 ##' @inheritParams orderly_metadata
+##' @inheritParams orderly_query
 ##'
 ##' @return A character vector of matching ids. In the case of no
 ##'   match from a query returning a single value (e.g., `latest(...)`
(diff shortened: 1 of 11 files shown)


R/query_search.R
```diff
@@ -53,34 +52,45 @@ orderly_search <- function(..., parameters = NULL, envir = parent.frame(),
...
-##'   [orderly2::orderly_dependency] you might pull a large
-##'   quantity of data.
+##' @param allow_remote Logical, indicating if we should allow packets
+##'   to be found that are not currently unpacked (i.e., are known
+##'   only to a location that we have metadata from). If this is
+##'   `TRUE`, then in conjunction with [orderly2::orderly_dependency]
+##'   you might pull a large quantity of data.  The default `NULL` is
+##'   `TRUE` if remote locations are listed explicitly as a character
+##'   vector in the `location` argument, or if you have specified
+##'   `pull_metadata = TRUE`, otherwise `FALSE`.
```
M-Kusumgar, line 62:
```suggestion
##'   you might pull a large quantity of data.  The default is `NULL`. This is
##'   `TRUE` if remote locations are listed explicitly as a character
##'   vector in the `location` argument, or if you have specified
##'   `pull_metadata = TRUE`, otherwise `FALSE`.
```

M-Kusumgar: think it reads a bit clearer?

M-Kusumgar (approved):
Small suggestion but yh i think this is in the right direction based on memory of conversation with jeff, the orderly search documentation improvement is also much clearer thanks

PR #187, opened 2024-10-18
Ergonomic improvements for querying, part 2

### LINKED ISSUE #156: orderly_metadata_extract(options = list(allow_remote = TRUE, location = "root2"))

richfitz:
This PR fixes #156 (mostly) by making the arguments to the whole search system much easier to discover.

All arguments now come through as top-level arguments, with the docs being easier to follow and autocomplete better.  Some defaults have been changed to better reflect what is wanted, and in other cases arguments that make no sense have been removed (e.g., if we would always want to include remote locations).  The old options interface is supported for now with a deprecation warning

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.45
+Version: 1.99.46
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
                     email = "rich.fitzjohn@gmail.com"),
              person("Robert", "Ashton", role = "aut"),
--- a/R/interactive.R
+++ b/R/interactive.R
@@ -82,9 +82,9 @@ detect_orderly_interactive_path <- function(
 orderly_interactive_set_search_options <- function(location = NULL,
                                                    allow_remote = NULL,
                                                    pull_metadata = FALSE) {
-  options <- orderly_search_options(location = location,
-                                    allow_remote = allow_remote,
-                                    pull_metadata = pull_metadata)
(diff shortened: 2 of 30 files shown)


M-Kusumgar (approved):
looks good to me, definitely a fan of the explicit arguments

PR #188, opened 2024-10-21
Verify location before addition

richfitz:
Merge after #185, contains those commits.

This PR adds new arguments `verify` and `quiet` to `orderly_location_add`:

* `verify` will try and connect to the location before addition, erroring (and therefore making no changes) if it is misconfigured
* `quiet` uses the same approach as #185 for controlling how chatty we are while we do this

This PR also fixes a small logic error in `save_token` where we say "Defaults to `TRUE` if `NULL`" - this looks to have been lost at some point and it remains as `NULL` until referenced within `if ()` at which point we threw an error.  Now set set `save_token` to the value of `is.null(token)`

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -114,10 +114,7 @@ orderly_location_add <- function(name, type, args, verify = TRUE, root = NULL) {
 
   if (verify) {
     cli_alert_info("Testing location")
-    driver <- location_driver_create(type, args)
-    if (!is.null(driver$authorise)) {
-      driver$authorise()
-    }
+    location_driver_create(type, args)$verify()
     cli_alert_success("Location configured successfully")
   }
 
--- a/R/location.R
+++ b/R/location.R
@@ -78,14 +78,20 @@
 ##' @param verify Logical, indicating if we should verify that the
 ##'   location can be used before adding.
 ##'
+##' @param quiet Logical, indicating if we should print information
+##'   while configuring and creating locations.  If not given, we use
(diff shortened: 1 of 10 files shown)


richfitz:
That coverage report from codecov is out of date and not being updated for its own mysterious reasons

R/location.R
```diff
@@ -75,26 +75,27 @@
...
 
   location_check_new_name(root, name, environment())
-  match_value(type, setdiff(location_types, location_reserved_name),
-              call = environment())
+  match_value(type, setdiff(location_types, location_reserved_name))
 
-  loc <- new_location_entry(name, type, args, call = environment())
   if (type == "path") {
     assert_scalar_character(args$path, name = "path")
-    root_open(args$path, require_orderly = FALSE)
```
plietar, line 97: As far as I can tell this means there's no more validation of path locations, since the location_path class does not have an `authorize` method.

plietar (approved):
I like this, especially since it means it will perform and cache authentication when adding the location, not randomly later on.

I would expect this to do a bit more than just calling `authorize()`. If you hit the root of the API there's a basic JSON document:
```r
dput(orderly_location_packit("https://packit.dide.ic.ac.uk/reside/")$client$request("/"))
#> list(status = "success", data = list(schema_version = "0.1.1"), errors = NULL)
```

Checking the status is good enough. This should work on both packit and outpack_server.

For a path location it should retain the existing `root_open` behaviour.

PR #189, opened 2024-10-22
Allow envvar 'ORDERLY_ROOT' to use alternative root by default

richfitz:
Add support for an environment variable `ORDERLY_ROOT` that we look up when `root` is not provided (i.e., is `NULL`).  This allows us to control the apparent root of orderly in documentation (see usage in the vignettes here), and appears more usable in quarto docs than the previous `setwd` hack, which is now removed.  This is not documented or advertised anywhere, but we can do do if we think it makes sense to allow users to work this way too.

### DIFF
--- a/vignettes/collaboration.Rmd
+++ b/vignettes/collaboration.Rmd
@@ -127,7 +127,12 @@ orderly2::orderly_list_src()
 Alice needs to run `orderly::orderly_init()` first:
 
 ```{r, as = "alice", orderly_root = path_alice}
-orderly2::orderly_init(".")
+orderly2::orderly_init(path_alice)
+```
+
+after which Alice can use orderly commands:
+
+```{r, orderly_root = path_alice}
 orderly2::orderly_list_src()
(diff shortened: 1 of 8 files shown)


vignettes/collaboration.Rmd
```diff
@@ -114,26 +114,31 @@ Both Alice and Bob clone this repo onto their machines, using their favourite gi
 
 At the moment, including hidden files (except `.git`), Alice sees:
 
-```{r, as = "alice", inwd = path_alice}
+```{r, as = "alice", orderly_root = path_alice}
 fs::dir_tree(".", all = TRUE, glob = ".git", invert = TRUE)
```
plietar, line 118: Wouldn't that affect the output of this block?

plietar (approved):

vignettes/collaboration.Rmd
richfitz: thanks, fixed that

PR #190, opened 2024-10-22
Indicate if extracted metadata is local or not

richfitz:
More from Jeff's request in #156 - an indication of metadata refers to a packet available locally.

### DIFF
--- a/R/outpack_tools.R
+++ b/R/outpack_tools.R
@@ -102,7 +102,11 @@
 ##' are more complex, these will be list columns.
 ##'
 ##' You must not provide `id`; it is always returned and always first
-##' as a character vector column.
+##' as a character vector column.  If your extraction could possibly
+##' return data from locations (i.e., you have `allow_remote = TRUE`
+##' or have given a value for `location`) then we add a logical column
+##' `local` which indicates if the packet is local to your archive,
+##' meaning that you have all the files from it locally.
 ##'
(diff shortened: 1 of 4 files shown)


R/outpack_tools.R
```diff
@@ -222,6 +226,9 @@ orderly_metadata_extract <- function(expr = NULL, name = NULL, location = NULL,
 
   envir <- environment()
   ret <- data_frame(id = ids)
+  if (isTRUE(allow_remote) || length(location) > 0) {
+    ret$local <- ids %in% root$index$unpacked()
+  }
```
M-Kusumgar, line 231: local boolean is good, but do we not have any way of just having a "location" column that has "local" as a special value and the name of any other location someone has specified? i can imagine users potentially wanting info about what remote location a packet has come from

M-Kusumgar (commented):
looks good, have a suggestion that if possible might be nicer

R/outpack_tools.R
richfitz: agree that this would be nicer....*but* as packets can be found at multiple locations it's awkward as we have to return a list column. If we think that's worth it it's very easy to add, but they're a bit of a pain to use imo (we do return them liberally elsewhere in this function though)

richfitz: OK, I think this is worthwhile and have done it

M-Kusumgar (approved):
thanks LGTM!

PR #191, opened 2024-10-22
Make example runner less clever

richfitz:
After a short fight, I can't get the withr deferred events to behave nicely under `devtools::run_examples`, which is I presume a similar environment to `pkgdown` - which leads to the silently failing examples on the docs.  I've updated this to be very simple now and not try and clean up automatically at all, which should mean that the docs now run with no problems.

### DIFF
--- a/R/cleanup.R
+++ b/R/cleanup.R
@@ -66,7 +66,7 @@
 ##' # Do the actual deletion:
 ##' orderly2::orderly_cleanup("data", root = path)
 ##'
-##' unlink(path, recursive = TRUE)
+##' fs::dir_delete(path, recursive = TRUE)
 orderly_cleanup <- function(name = NULL, dry_run = FALSE, root = NULL) {
   status <- orderly_cleanup_status(name, root)
   n <- length(status$delete)
--- a/R/cleanup.R
+++ b/R/cleanup.R
@@ -42,7 +42,6 @@
 ##'
 ##' @export
 ##' @examples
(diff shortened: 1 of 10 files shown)


R/cleanup.R
```diff
@@ -42,6 +42,7 @@
 ##'
 ##' @export
 ##' @examples
+##' message(sprintf("Running in %s", format(environment())))
```
plietar, line 45: Accidental leftover?

plietar (approved):
I would use `fs::dir_delete`. I hate that unlink just fails silently.

It's a bit of a shame the `r-lib` packages don't handle this well. Might be worth raising the issue with pkgload/withr/

pkgload does a `new.environment`, and `withr` uses `do.call` to run `on.exit` in that environment. I guess no one exits that environment, and I don't even know if there's an API for doing that (didn't find anything in the rlang package). Maybe if `pkgload` had a `do.call(return, list(), env=env)` line it would work?

https://github.com/r-lib/pkgload/blob/96d1a746f12ebf0a54157ed2a256bdc32358f9f6/R/dev-example.R#L38

https://github.com/r-lib/withr/blob/82daa902b654edabdd523e09d9a3b1b7216f7cc3/R/defer.R#L108-L111

R/cleanup.R
richfitz: indeed, trying to find out what withr/devtools were thinking they had as their environment - turns out it was exactly what I expected them to have, but it just did not work

PR #192, opened 2024-10-23
Cope better with relative locations

richfitz:
Previously we allowed relative paths to be used for locations but it was ambiguous where they were looked up.  This is a problem in contexts where absolute paths are not suitable (e.g., on the cluster where the mount point on the nodes is different to the submitting workstation).

This PR removes the ambiguity - we take the path relative to the root, and **not** relative to the command that calls `orderly_location_add_path()`.  This might be confusing so there's a special case error that we throw if it looks like the wrong relative path was chosen.

When we load the driver we change back to the root and then interpret the path at that point.  This is also done when the user verifies addition (in #188).

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -96,6 +96,21 @@ orderly_location_add <- function(name, type, args, verify = TRUE, root = NULL) {
 
   if (type == "path") {
     assert_scalar_character(args$path, name = "path")
+    if (!fs::is_absolute_path(args$path)) {
+      ## This only happens where the current working directory is not
+      ## the same as the root.
+      root_error <- !file.exists(file.path(root$path, args$path)) &&
+        file.exists(args$path)
+      if (root_error) {
+        root_fix <- as.character(fs::path_rel(args$path, root$path))
+        cli::cli_abort(
+          c("'path' must be given relative to the orderly root",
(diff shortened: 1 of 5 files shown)


R/location.R
```diff
@@ -96,6 +96,21 @@ orderly_location_add <- function(name, type, args, verify = TRUE, root = NULL) {
...
+        file.exists(args$path)
+      if (root_error) {
+        root_fix <- as.character(fs::path_rel(args$path, root$path))
+        cli::cli_abort(
+          c("'path' must be given relative to the orderly root",
+            x = paste("You have provided the relative path '{args$path}',",
+                      "which exists, but does not exist relatively to",
+                      "'{root$path}', the root of your orderly archive"),
+            i = "Consider passing '{root_fix}' instead"))
+      }
```
plietar, line 112: This is really nice diagnostics! Really solves the problem of having to choose between two equally acceptable behaviours.

plietar (approved):
I don’t love the idea of using the current working directory as some kind of implicit argument to the location_path constructor, but I guess I can live with it.

Aren’t we glad R is single threaded and we don’t have to worry about shared mutable state.

Issue #155, opened 2024-07-04
`orderly_location_push()` by query / consistency of args with `orderly_location_pull_packet()`


jeffeaton:
* `orderly_location_push()` requires specification of a `packet_id = `, but `orderly_location_pull_packet()` gives a flexible interface to pull by query.

* The argument for `orderly_location_push()` is `packet_id`, but that argument does not exist for `orderly_location_pull_packet()` and throws an error:
```
> orderly_location_pull_packet(packet_id = "20240704-215806-90fbb966")
Error in orderly_query(expr, ...) : 
  unused argument (packet_id = "20240704-215806-90fbb966")
```
(`orderly_location_pull_packet("20240704-215806-90fbb966")` works, but I can't actually figure out what the argument name that matches..)

`orderly_location_push()` requires specification of a `packet_id = `,

From using, it feels like these should be similar to each other. 

1. Can `orderly_location_push()` accept same query format?
2. Can `orderly_location_pull_packet()` and `orderly_location_push()` use the same argument when specifying a `packet_id`?

Similar to #146, it feels like the function names `orderly_location_push()` and `orderly_location_pull_packet()` should mirror each other. For me I favour shortening `orderly_location_pull_packet()` to `orderly_location_pull()` :).

richfitz:
Fixed in #193

Issue #153, opened 2024-07-04
automatically update metadata in `orderly_location_push()`?


jeffeaton:
From working through pushing / pulling w/ collaborators today: it is counterintuitive that when I push a packet, it does not immediately appear in `orderly_extract_metadata()`

Would it be possible for `orderly_location_push()` to also update the metadata for the location?

richfitz:
Fixed in #193

Issue #143, opened 2024-06-18
chattier orderly_location_push() and clarity about data vs. metadata in location archive


jeffeaton:
The flexibility around pushing / pulling packets in orderly2 is great. But it has left me somewhat confused about when data is being pushed around vs. only metadata.

Example: I ran several orderly reports in sequences, and then pushed the final one to a remote location. In orderly1 `orderly_push_archive()` would give me a summary of each of the dependency tasks that it was pushing before my final task.

Now when I call `orderly_location_push()`, it simply runs and returns silently:
```
> orderly2::orderly_location_push("20240618-200013-3a767016", "hiv-inference")
> 
```

This has left me unsure whether this pushed only the contents of this final task + metadata for the upstream tasks, or it also pushed the data underpinning the full dependency tree. My location (`"hiv-inference"`) was created using `orderly_init(..., require_complete_tree = TRUE)`, so I presumed it is pushing all of the data for upstream tasks; but I'm not sure how to verify that.

I think it would be helpful for `orderly2::orderly_location_push()` to (perhaps optionally) list the upstream packets, and whether they are pushed or already on the remote location.

richfitz:
Fixed in #193

PR #193, opened 2024-10-23
Gh 143

richfitz:
A collection of improvements for push

* make it chattier, fixing #143 
* allow a query, fixing #155 
* throw a sensible error if pushing a packet which does not exist (previously just an assertion that was incomprehensible to the user)
* add a `dry_run` argument which informs about what would be done but makes no changes
* pull updated metadata after pushing, fixing #153

### DIFF
--- a/R/location.R
+++ b/R/location.R
@@ -556,7 +556,7 @@ location_push_files <- function(files, driver, root) {
             i = paste("The original file has been changed or deleted.",
                       "Details are above")))
       }
-      size <- as.character(fs::file_size(src))
+      size <- pretty_bytes(fs::file_size(src))
       cli_progress_update()
       driver$push_file(src, hash)
     }
--- a/R/location.R
+++ b/R/location.R
@@ -493,9 +493,37 @@ orderly_location_push <- function(packet_id, location, root = NULL) {
                                           environment())
   plan <- location_build_push_plan(packet_id, location_name, root)
 
-  if (length(plan$files) > 0 || length(plan$packet_id) > 0) {
+  if (length(plan$files) == 0 && length(plan$packet_id) == 0) {
+    cli_alert_success("Nothing to push, everything up to date")
+  } else {
+    cli_alert_info(
+      paste("Pushing {length(plan$files)} file{?s} for",
+            "{length(plan$packet_id)} packet{?s}"))
     driver <- location_driver(location_name, root)
-    for (hash in plan$files) {
+    location_push_files(plan$files, driver, root)
+    location_push_metadata(plan$packet_id, driver, root)
+  }
+
+  invisible(plan)
+}
+
+
+location_push_files <- function(files, driver, root) {
+  n_files <- length(files)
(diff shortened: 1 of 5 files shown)


R/location.R
```diff
@@ -447,48 +447,115 @@ orderly_location_pull_packet <- function(expr,
...
+  }
+
+  plan <- location_build_push_plan(ids, location_name, root)
+
+  if (length(plan$files) == 0 && length(plan$packet_id) == 0) {
+    cli_alert_success("Nothing to push, everything up to date")
+  } else {
+    cli_alert_info(
+      paste("Pushing {length(plan$files)} file{?s} for",
+            "{length(plan$packet_id)} packet{?s}"))
```
M-Kusumgar, line 497: is just length too little info? wonder if we can list these out with a ... at least, better yet list out an abbreviated version and let them expand if they wish

R/location.R
```diff
@@ -447,48 +447,115 @@ orderly_location_pull_packet <- function(expr,
...
+    if (dry_run) {
+      cli_alert_info("Not making any changes, as 'dry_run = TRUE'")
+    } else {
+      driver <- location_driver(location_name, root)
+      location_push_files(plan$files, driver, root)
+      location_push_metadata(plan$packet_id, driver, root)
+      orderly_location_pull_metadata(location_name, root)
```
M-Kusumgar, line 504: yh definitely think this is a sensible default to have behaviour wise

tests/testthat/test-location-path.R
```diff
@@ -356,6 +356,68 @@ test_that("Fail to push sensibly if files have been changed", {
...
+  client <- create_temporary_root()
+  id1 <- create_random_packet(client, parameters = list(a = 1))
+  id2 <- create_random_packet(client, parameters = list(a = 2))
+  id3 <- create_random_packet(client, parameters = list(a = 1))
+
+  server <- create_temporary_root(use_file_store = TRUE, path_archive = NULL)
+  orderly_location_add_path("server", path = server$path, root = client)
+
+  plan <- orderly_location_push("parameter:a == 1", "server",
+                                dry_run = TRUE, root = client)
```
M-Kusumgar, line 374: perhaps an expected message test here?

M-Kusumgar (commented):
Nice, this is pretty good, got a couple comments here and there

R/location.R
richfitz: We don't actually have filenames at this point because of the 1:many correspondence of hash to file.  All we hold there is hashes and ids neither of which are terribly useful

R/location.R
richfitz: Seems obvious in retrospect - I actually missed this suggestion in Jeff's pile-on at first :)

tests/testthat/test-location-path.R
richfitz: sure I can add that

M-Kusumgar (approved):
Nice one

PR #194, opened 2024-10-23
Rename "pull" functions

### LINKED ISSUE #146: suggestion: change function name `orderly_location_pull_metadata()` to `orderly_location_metadata_update()`
As I work with orderly2, I am finding the similarity of function names `orderly_location_pull_metadata()` and `orderly_location_pull_packet()` a bit confusing for myself because they sound quite similar, but do quite different things and have different signatures.

What about changing the name `orderly_location_pull_metadata()` to `orderly_location_metadata_update()` or similar? To me, 'update' captures what the function is doing more intuitively than 'pulling'.

If that change is adopted, `orderly_location_pull_packet()` could potentially be abbreviated to simply `orderly_location_pull()`.

richfitz:
This PR fixes #146, by renaming the location pull commands:

* `orderly_location_pull_metadata` becomes `orderly_location_metadata_fetch`
* `orderly_location_pull_packet` becomes `orderly_location_pull`

There's a bit of deprecation support, but I think that won't be needed for long.

Two small issues arising:

* Do we prefer `metadata_fetch` or `fetch_metadata`?
* The `pull_metadata` argument now does not make much sense; fixing this is a little more disruptive but we are uniquely placed to fix it after #187 as people are updating anyway


M-Kusumgar (approved):
awesome, i think i actually prefer `orderly_location_fetch_metadata` after youve brought it up, reads more like english

yh `pull_metadata` arg should probably be `fetch_metadata`, random thought, might be cool to replicate docker style pull arg so instead of boolean, we can have "always", "never" and "missing" so we automatically pull in missing ones on demand

richfitz:
> yh pull_metadata arg should probably be fetch_metadata, random thought, might be cool to replicate docker style pull arg so instead of boolean, we can have "always", "never" and "missing" so we automatically pull in missing ones on demand

I've done the rename.  The ternery arg here is probably not really the same as what we do though so I've left that as boolean but we can discuss that later if you think there's something I'm missing here

PR #201, opened 2024-11-29
Add documenting comments to orderly.json schema

david-mears-2:
To clarify the intent behind the properties/what terms refer to

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly2
 Title: Orderly Next Generation
-Version: 1.99.58
+Version: 1.99.59
 Authors@R: c(person("Rich", "FitzJohn", role = c("aut", "cre"),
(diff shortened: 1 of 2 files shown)


david-mears-2:
Unfortunately the 'windows-latest' R-CMD-check fails. I re-ran it and it failed again.

plietar:
@richfitz how is this schema coordinated across repos (if at all)?

We have https://github.com/mrc-ide/outpack for the outpack schemas, but I can't find any equivalent for `orderly.json`.

plietar:
I couldn't reproduce the build failure on a Windows VM :disappointed:

richfitz:
The idea was that orderly2 is responsible for its schemas, it would hold them itself.  It imports schemas from outpack (there's a script somewhere to pull them in)

richfitz (approved):

PR #209, opened 2025-05-02
Add parameters that are less magical

richfitz:
Quite a bit of change to support this, but the changes to the examples should highlight the intent. Rather than writing:

```
orderly_parameters(foo = NULL, bar = NULL)
```

and then, during `orderly_run()` writing `foo` and `bar` to the execution environment, we now support and recommend:

```
pars <- orderly_parameters(foo = NULL, bar = NULL)
```

With this approach `foo` and `bar` are accessed as `pars$foo` and `pars$bar`.  The name used on the lhs is whatever is wanted by the user. 

That part of the change is not that bad but where it gets hairy:

* supporting the previous version with a warning
* working in interactive mode (this is likely to require some fine-tuning)


tests/testthat/examples/depends-query/depends-query.R
```diff
@@ -1,6 +1,6 @@
-orderly2::orderly_parameters(a = NULL, b = NULL, c = NULL)
+pars <- orderly2::orderly_parameters(a = NULL, b = NULL, c = NULL)
 
-value <- c
```
weshinsley, line 3: I am curious about how `pars$a`, and `pars$b` can be referenced as `this:a` and `this:b` on lines 6 and 7 - how does it know it's `pars` ? (Is this residue-magic?)

weshinsley (commented):
Ok - I think I mostly get this, although the interactive mode/how it works has faded out of my memory a bit. 

A few doc typos and a question or two below - then vignettes/dependencies.Rmd and introduction.Rmd I think need updating.

tests/testthat/examples/depends-query/depends-query.R
richfitz: this is magic that remains - the orderly report "knows" what its parameters are - the only change here is how user code finds that out

weshinsley (approved):

PR #221, opened 2025-07-28
Drop device flow httr2 browser workaround

richfitz:
https://github.com/r-lib/httr2/pull/763 is released to CRAN now, so this PR drops the function that allowed control over browser opening, and uses the support now present in httr2.

httr2 now redacts Authorization headers by turning them into weakrefs, so I've adapted the tests to check that these are now as expected as we can't check against strings anymore.


plietar (approved):
I am so confused about why httr2 people chose to use a weakref to represent redacted headers, but that's not our problem I guess.

PR #226, opened 2025-08-15
Rename orderly2 to orderly

richfitz:
A big find-and-replace job.  This was mechanical and there might be a few missed odd sentences here and there as a result.

This merges into the epic branch (#225) and not main (yet)


vignettes/migrating.Rmd
```diff
@@ -141,4 +141,4 @@ install.packages(
 
 and you can work with the sources or outputs as you wish.  You will need to replace any namespaced calls `orderly::fn` with `orderly1::fn` (these might appear in your documentation, helper scripts, or similar).
 
-If you try to use `orderly2` with version 1 sources, then you will get an error message that will direct you to this document to decide the best way forward.
+If you try to use `orderly` with version 1 sources, then you will get an error message that will direct you to this document to decide the best way forward.
```
weshinsley, line 144: Could insert a (version 2) or something here?

weshinsley (approved):
Looks good - a couple of places where I wondered whether we need to say "since version 2" or something - but I'm in two minds; that will quickly become unnecessary and perhaps it already is.

PR #230, opened 2025-09-03
Add ability to migrate orderly2 sources to orderly

richfitz:
A very simple-minded source migration tool for people using `orderly2`; just does a find and replace!  Guidance and more helpers coming in the next PR


weshinsley (approved):
Looks good to me!

PR #235, opened 2025-09-12
Skip some platform-dependent tests on CRAN

richfitz:
These are breaking the universe build, and would likely also be problematic on CRAN


richfitz:
Closed as combined with #233

PR #240, opened 2025-09-26
Refactor root detection logic

richfitz:
I want to replace the yaml configuration, but I do not want to make the existing overcomplicated logic around root detection more complicated.  This PR does some preliminary refactoring, with the aim of reducing the number of references to `orderly_config.yml` that remain; the idea is that with this I can write a config reader that accepts whatever format we need to, and that I can then count either the json or yml as the root marker file.

I've preserved as much functionality as I could see we were using, though a few error messages will have changed very slightly

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: orderly
 Title: Lightweight Reproducible Reporting
-Version: 1.99.88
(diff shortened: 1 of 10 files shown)


weshinsley (approved):
Ok - had to stare at it a bit!

PR #242, opened 2025-10-03
Don't try to change any file in a previously run packet

richfitz:
First, I tried just excluding the files: that works quite well (first commit).  Then I figured that actually we should just prevent this entirely because it's not really safe.


weshinsley (approved):
All clear!

PR #220, opened 2025-07-28
Add more examples

richfitz:
~Merge after #221, contains those commits~

This PR goes through and adds a bunch of examples to the orderly function help.  There are a few I've not yet done - the plugins (which have their own vignette), interactive search option setting (which requires some changes I think), ~and some of the query/search bits which feel like they're going to take a little more effort~.

I've overhauled the example interface a bit to make this work well, and put in new support showing the contents of an orderly file so that we can have something to show for the metadata examples.

I also have snuck in a small doc improvement for the migration guide

PR #243, opened 2025-10-06
Add final examples

richfitz:
These were missed in #220, but are not hard.  I had wondered about changing the interface for setting the orderly interactive options, but it's complicated enough already so have left it as is. The query index interface is internal and there's no need to generate Rd files for it (and I can never remember how grumpy CRAN get about this)


weshinsley (approved):
Just one ~cornetto~ typo

PR #245, opened 2025-10-07
Add logo

richfitz:
Courtesy of @kgaythorpe


weshinsley (approved):
Majestic. I've merged the vignette one before this, so think we just need a version bump?

PR #246, opened 2025-10-08
Retire buildkite docker build

richfitz:
We don't use this anywhere and this makes it clear that the docker build and bk task are not important once it comes time to retire the dockerhub and buildkite services


EmmaLRussell (commented):
So, orderly now just has a gha ci pipeline, not buildkite?

And there was previously an orderly docker image, but now orderly is being pulled into other images instead (like orderly runner)..?

